PR #2, opened 2023-10-12
Add scripts for signing merkle trees

alikonuk1:
closes [api3dao/tasks#194](https://github.com/api3dao/tasks/issues/194)


scripts/merkle/dapiManagementMT.js
```diff
@@ -0,0 +1,51 @@
+const fs = require('fs');
+const path = require('path');
+const { signEIP712Message, constructMerkleTree } = require('./merkle-utils');
+const ethers = require('ethers');
+
+// Reading the input data
+const inputData = JSON.parse(fs.readFileSync(path.join(__dirname, 'treeValues', 'dapiManagementMT.json'), 'utf8'));
+const values = inputData.values;
+
+const tree = constructMerkleTree(values, ['string', 'bytes32', 'address']);
```
acenolaza, line 10: While working on `DapiDataRegistry` I started thinking that is nice to have `dapiName: 'API3/USD'` string in the dapiManagementMT.json metadata file but shouldn't we construct the merkle tree with the `bytes32` string instead since even the `DapiServer.sol` is expecting to received the formatted `bytes32` string when calling `SetDapiName()`?
```suggestion
const tree = constructMerkleTree(values, ['bytes32', 'bytes32', 'address']);
```

bbenligiray: Yes, the contract wouldn't be able to convert `string` to `bytes32` aynway

acenolaza: I guess it is now possible to do something like the following to convert it but it would consume more gas as well:
`api3ServerV1.setDapiName(bytes32(abi.encodePacked(dapiName)), dataFeedId);`

bbenligiray (changes requested):
Thanks, this is generally what is asked in the issue

A few requests:
- The signing scripts have a lot of code duplication between them. Can you implement them as a single script that works according to the hash type?
- I can't tell what the fields are in the JSON files. Can you document in the README how we should be using them?
- Also, can you add the file that keeps the list of root signers? For now you can populate it with the default hardhat signers.

package.json
```diff
@@ -46,5 +50,10 @@
     "ts-node": ">=8.0.0",
     "typechain": "^8.1.0",
     "typescript": ">=4.5.0"
+  },
+  "dependencies": {
+    "@api3/airnode-protocol-v1": "^2.10.0",
+    "@openzeppelin/contracts": "4.9.3",
```
bbenligiray, line 56: Shouldn't need these

scripts/merkle/treeValues/apiIntegrationMT.json
```diff
@@ -0,0 +1,15 @@
+{
```
bbenligiray, line 1: Move these JSON files (and the list of root signers) in `data/` at the repo root (a bit like https://github.com/api3dao/manager-multisig/tree/main/chain/fallback). These files will keep the actual values when we go to production, may as well set it up now.

scripts/merkle/dapiManagementMT.js
bbenligiray: `bytes32(abi.encodePacked(stringVariable))` truncates the string down to 32 characters so that's not really an option imo

scripts/signMerkleTree.js
```diff
@@ -0,0 +1,107 @@
...
+require('dotenv').config();
+
+const mnemonic = process.env.MNEMONIC;
+const wallet = ethers.Wallet.fromMnemonic(mnemonic);
+
+// Define the domain structure for EIP-712 signing
+const domain = {
+  name: 'HashRegistry',
+  version: '1.0.0',
+  chainId: 1,
```
acenolaza, line 14: Since this script is not run using hardhat then I guess the `chainId` should come as command line arg, right?

scripts/signMerkleTree.js
```diff
@@ -0,0 +1,107 @@
...
+  verifyingContract: '0x0000000000000000000000000000000000000000',
```
acenolaza, line 15: This can be read from `references.json` (to be added I guess)

README.md
```diff
@@ -9,3 +9,93 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+  }
+}
+```
+
+Each Merkle tree inside the `merkleTrees` object should have a timestamp, values array, merkleRoot, and signatures object.
+
+2. Run the script with the desired Merkle tree name:
+
+```bash
+yarn sign [MerkleTreeName]
```
acenolaza, line 54: Since `MerkleTreeName` becomes the `hashType` should we document where to get them from? [I've added 2 of the 4](https://github.com/api3dao/dapi-management/blob/169-DapiDataRegistry/contracts/DapiDataRegistry.sol#L25-L28) to DapiDataRegistry (price I'm not sure and fallback should be added to DapiFallbackV2 I guess)

scripts/signMerkleTree.js
```diff
@@ -0,0 +1,107 @@
...
+  return StandardMerkleTree.of(values, dataTypes);
+}
+
+// Predefined data types for each Merkle tree
+const MERKLE_TREE_DATATYPES = {
+  priceMT: ['bytes32', 'uint256', 'bytes32', 'uint256', 'uint256'],
+  dapiManagementMT: ['bytes32', 'bytes32', 'address'],
+  dapiFallbackMT: ['bytes32', 'bytes32', 'address'],
+  apiIntegrationMT: ['address', 'bytes32', 'bytes32'],
+};
```
acenolaza, line 50: I've just now realized that the [descriptions](https://github.com/api3dao/dapi-management/blob/169-DapiDataRegistry/contracts/DapiDataRegistry.sol#L25-L28) I used in DapiDataRegistry do not match the ones used as keys in here 😁

scripts/signMerkleTree.js
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  const signature = await wallet._signTypedData(domain, types, message);
+  return signature;
+}
+
+// Function to construct a Merkle tree from provided values and data types
+function constructMerkleTree(values, dataTypes) {
+  return StandardMerkleTree.of(values, dataTypes);
+}
```
acenolaza, line 42: These functions are probably going to be handy for frontend signing as well, right? I wonder if we should  export them from their own module. signEIP712Message might need to receive a PK as arg instead of relying on a global `wallet` object

README.md
```diff
@@ -9,3 +9,93 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+  "rootSigners": [
+    "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+    ...
+  ],
+  "merkleTrees": {
+    "priceMT": {
+      "timestamp": 1676940000,
+      "values": [...],
+      "merkleRoot": "",
+      "signatures": {}
```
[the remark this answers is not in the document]
alikonuk1: Yeah so conclusion was keeping it this way if I remember correctly 😄

README.md
```diff
@@ -9,3 +9,93 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+## Usage
+
+1. Update the `metadata.json` file located in the `data` directory with the required values for your Merkle tree. The structure of `metadata.json` should look like this:
+
+```json
+{
+  "rootSigners": [
+    "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+    ...
+  ],
```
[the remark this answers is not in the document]
alikonuk1: Oh good catch. This makes me want to divide json files by their hashType's, would be seen tidier maybe?

scripts/signMerkleTree.js
alikonuk1: Yes it will be added when we have `references.json`. I think we will have to wait for deploy scripts for it?

README.md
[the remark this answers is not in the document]
acenolaza: Or you can just have a `rootSigners` array within each merkle tree type (right above timestamp for each type)

scripts/signMerkleTree.js
acenolaza: Yes, I guess those will be added [here](https://github.com/api3dao/tasks/issues/196)

README.md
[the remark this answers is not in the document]
bbenligiray: Each MT type will have its own directory with a `metadata.json` file each. This will go in there. Alternatively, just have a `root-signers.json` file, as that is the only metadata field.

README.md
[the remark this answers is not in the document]
bbenligiray: It's a good idea to bundle the signatures with the tree values considering that there will be two sets of tree values (`current` and `previous`)

README.md
```diff
@@ -9,3 +9,102 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+  "merkleTrees": {
+    "priceMT": {
```
bbenligiray, line 38: I'm seeing `priceMT`, `Price Merkle Tree`, `Price merkle tree root` in the PR. Pick a name for each and use them consistently.

README.md
```diff
@@ -9,3 +9,102 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+      "timestamp": 1676940000,
+      "values": [...],
+      "merkleRoot": "",
```
bbenligiray, line 41: No need for the root

README.md
```diff
@@ -9,3 +9,102 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+- **Values**:
+  - **dAPI Name (bytes32)**: The name of the dAPI, hashed.
+  - **Nodary Beacon ID (bytes32)**: Identifier for the Nodary Beacon.
+  - **Nodary Sponsor Wallet Address (address)**: Wallet address of the Nodary sponsor wallet.
+
+### 4. API Integration Merkle Tree
+
+- **Values**:
+  - **Airnode Address (address)**: The contract address of the Airnode.
+  - **Signed API URL (bytes32)**: The signed URL of the API, hashed. Verified by referring to the DNS records of the base URL.
```
bbenligiray, line 98: This is not hashed, it's the full URL as a `string`

README.md
```diff
@@ -9,3 +9,102 @@ This contract is intended to be used as a generic hash registry. Users of this c
...
+  - **dAPI Name (bytes32)**: The name of the dAPI, hashed.
+  - **Beacon Set ID (bytes32)**: Identifier for the beacon set.
+  - **dAPI Sponsor Wallet Address (address)**: Wallet address of the dAPI sponsor wallet.
+
+### 3. dAPI Fallback Merkle Tree
+
+- **Values**:
+  - **dAPI Name (bytes32)**: The name of the dAPI, hashed.
+  - **Nodary Beacon ID (bytes32)**: Identifier for the Nodary Beacon.
+  - **Nodary Sponsor Wallet Address (address)**: Wallet address of the Nodary sponsor wallet.
```
bbenligiray, line 92: There is nothing Nodary-specific about the schema of the Merkle tree. Any other data feed ID would work here.

data/metadata.json
```diff
@@ -0,0 +1,86 @@
...
+    "0x90F79bf6EB2c4f870365E785982E1f101E93b906",
+    "0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65",
+    "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+    "0x976EA74026E726554dB657fA54763abd0C3a0aa9",
+    "0x14dC79964da2C08b23698B3D3cc7Ca32193d9955",
+    "0x23618e81E3f5cdF7f54C3d65f7FBc0aBf5B21E8f",
+    "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720"
+  ],
+  "merkleTrees": {
+    "Price merkle tree root": {
```
bbenligiray, line 15: If the parent key is `merkleTrees`, you can't name this `Price merkle tree root` (a tree is not a root)

bbenligiray (changes requested):
The Merkle tree names:
- ~Price Merkle tree~ -> dAPI pricing Merkle tree
- dAPI management Merkle tree
- dAPI fallback Merkle tree
- ~API integration Merkle tree~ -> Signed API URL Merkle tree

bbenligiray:
```
data/
├─ dapi-pricing-merkle-tree-root/
│  ├─ hash-signers.json
│  ├─ current-hash.json
├─ dapi-management-merkle-tree-root/
│  ├─ hash-signers.json
│  ├─ current-hash.json
│  ├─ previous-hash.json
├─ dapi-fallback-merkle-tree-root/
│  ├─ hash-signers.json
│  ├─ current-hash.json
├─ signed-api-url-merkle-tree/
│  ├─ hash-signers.json
│  ├─ package.json
```

README.md
bbenligiray: Actually we can do something like this
```
{
  "timestamp": 1676940000,
  "hash": "0x1234...",
  "signatures": {
  },
  "merkleTreeValues": {
  }
}
```
where `merkleTreeValues` is optional. This can then support arbitrary hash types (not only Merkle tree roots) in the future.

scripts/signMerkleTree.js
```diff
@@ -0,0 +1,124 @@
...
+  console.error('You must provide a valid chainId as an argument!');
+  process.exit(1);
+}
+const chainId = parseInt(chainIdArg);
+
+// Define the domain structure for EIP-712 signing
+const domain = {
+  name: 'HashRegistry',
+  version: '1.0.0',
+  chainId: chainId,
```
bbenligiray, line 38: This implies that we'll need to sign all Merkle tree roots once per each chain, which wasn't the plan. We can hardcode a chain ID here, but EIP712.sol gets its chain ID from the chain, so it wouldn't work.

I guess leave this as is, but the contract and these scripts will have to be updated later to use ERC191 instead.

bbenligiray:
Apply the changes above and ask for another review (DM on Slack) and let's merge it then because this remained open for far too long. If I want any other changes I can open a PR of my own.

bbenligiray (approved):
You should commit the fixes individually, I won't bother checking this

PR #3, opened 2023-10-13
WIP: Adds DapiDataRegistry.sol contract

acenolaza:
This contract is still a WIP but I have reached the point where I think I have covered most of the requirements for this contract and I've made quite a few assumptions that need to be checked. I've also have a couple of questions and I think is better to just add a TODO in the code instead of asking privately over slack. The main goal of this PR is getting feedback as early as possible while I keep adding tests 🙏🏻


contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    /// @notice Register data role description
+    // string public constant override REGISTER_DATA_ROLE_DESCRIPTION = "dAPI data setter role";
+    string public constant REGISTER_DATA_ROLE_DESCRIPTION =
+        "Register dAPI data role";
+
+    /// @notice Register dAPI data role
+    // bytes32 public immutable override registerDataRole;
+    bytes32 public immutable registerDataRole;
```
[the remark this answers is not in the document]
bbenligiray: The convention is for the role description variable name, role description and the role variable name to all be the same. "Role" doesn't go in the description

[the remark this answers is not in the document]
bbenligiray: Also proposing registrar as a short and descriptive alternative

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+
+    modifier onlyRegisterDataRole() {
```
bbenligiray, line 28:
```suggestion
    modifier onlyRegistrar() {
```

bbenligiray: What I mean is when a modifier requires the sender to be a specific account, the convention is to call the modifier something like `onlyOwner`. `-Role` shouldn't go in here.

bbenligiray: As another note, I avoid using this `onlyX` naming convention for role checking purposes because it's inaccurate, the manager can also call these even if it doesn't have the role. I used to do
```
    modifier onlyRegistrarOrManager() {
```
instead but I couldn't find an example of that

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+
+    struct Beacon {
+        address airnode;
+        bytes32 templateId;
+    }
+    struct SignedApi {
+        bytes32 oisTitle;
+        string url;
+    }
+    // TODO: use uint128 to pack both in a single 32 bytes slot?
```
bbenligiray, line 67: Maybe

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+    struct UpdateParameters {
+        uint256 deviationThreshold;
+        uint256 heartbeatInterval;
+        // int224 deviationReference; Currently not used by Airseeker v1
```
bbenligiray, line 71: Not having this when we need it will be a big problem

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+
+    // This is updated using the API management merkle tree
+    mapping(address => SignedApi) public airnodeToSignedApi;
+
+    // The value should be a single value or an array of them
+    // This needs to be encoded so we can determine if it's a beacon
+    // or a beaconSet based on the lenght
+    // It can be udpated by anyone because the contract will hash the data and derive it
+    // Airseeker will need to multicall to read all data using a single RPC call
+    mapping(bytes32 => bytes) public dataFeedIdToDataFeedData;
```
bbenligiray, line 82:
```suggestion
    mapping(bytes32 => bytes) public dataFeedIdToData;
```

bbenligiray: In this `to` convention, I omit the duplicated subject

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+        registerDataRole = _deriveRole(
+            _deriveAdminRole(manager),
+            REGISTER_DATA_ROLE_DESCRIPTION
+        );
+        timestampedHashRegistry = _timestampedHashRegistry;
+        api3ServerV1 = _api3ServerV1;
+    }
+
+    function registerAirnodeSignedApiUrl(
+        bytes32 hashType,
```
bbenligiray, line 112: This function should know what hash type it uses, it shouldn't ask the user

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+        uint256 deviationThreshold,
+        uint256 heartbeatInterval
+    );
+
+    struct Beacon {
+        address airnode;
+        bytes32 templateId;
+    }
+    struct SignedApi {
+        bytes32 oisTitle;
```
bbenligiray, line 64: Actually let's leave out `oisTitle` from here. If needed it can be a separate MT.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+            bytes.concat(keccak256(abi.encode(airnode, oisTitle, url)))
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+
+        airnodeToSignedApi[airnode] = SignedApi(oisTitle, url);
+
+        emit RegisteredSignedApiUrl(airnode, oisTitle, url);
+    }
+
+    // TODO1: Should anyone really be able to call this?
```
bbenligiray, line 136: Yes

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+    // TODO2: Wouldn't it be simpler to register using an array of Beacon[]?
+    //        It would have a single element when beacon and multiple when beaconSet
```
bbenligiray, line 138: A single Beacon Beacon set and a single Beacon are different data feeds with different IDs (even though there is no good reason to use a single Beacon Beacon set). I didn't like the ambiguity.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+    function registerDatafeed(
+        bytes calldata dataFeedData
+    ) external returns (bytes32 dataFeedId) {
+        require(dataFeedData.length > 0, "Data feed data is empty");
+        // console.log(dataFeedData.length);
+        if (dataFeedData.length == 64) {
+            // DataFeedId maps to a beacon
+            Beacon memory beacon = abi.decode(dataFeedData, (Beacon));
```
bbenligiray, line 146: `Beacon` and `beacon` feel to vague as variable names to me. It can be the IDs or the values. Also not a fan of encoding-decoding structs, too fancy.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+            // console.log(beacon.airnode);
+            // console.logBytes32(beacon.templateId);
+            // Derive beacon ID
+            dataFeedId = keccak256(
+                abi.encode(beacon.airnode, beacon.templateId)
+            );
+        } else {
+            // DataFeedId maps to a beaconSet
+            bytes[] memory beacons = abi.decode(dataFeedData, (bytes[]));
```
bbenligiray, line 155: Why not decode into `(address[], bytes32[])`?

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+                // Derive beacon ID
+                beaconIds[ind] = keccak256(
+                    abi.encode(beacon.airnode, beacon.templateId)
+                );
+            }
+            // Derive beacon set ID
+            dataFeedId = keccak256(abi.encode(beaconIds));
+        }
+
+        dataFeedIdToDataFeedData[dataFeedId] = dataFeedData;
```
bbenligiray, line 170: Here's an interesting attack, I get the data feed data of a dAPI, add a huge bytes string at the end and re-register it, and the Airseeker stops being able to fetch it from the chain as a result. It feels like we should validate that data feed data should be exact/have no trailing data.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+
+        emit RegisteredDataFeed(dataFeedId, dataFeedData);
+    }
+
+    function registerDapi(
+        bytes32 hashType,
+        bytes32 dapiName,
+        bytes32 dataFeedId,
+        address sponsorWallet,
+        uint256 deviationThreshold, // TODO: how to represent this? basis points? 0.25% == 25
```
bbenligiray, line 180: Refer to https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/extensions/BeaconSetUpdatesWithPsp.sol#L111 for the time being

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+            dataFeedId = keccak256(abi.encode(beaconIds));
+        }
+
+        dataFeedIdToDataFeedData[dataFeedId] = dataFeedData;
+
+        emit RegisteredDataFeed(dataFeedId, dataFeedData);
+    }
+
+    function registerDapi(
+        bytes32 hashType,
```
bbenligiray, line 176: Shouldn't ask the hash type

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+        emit RegisteredDapi(
+            dapiName,
+            dataFeedId,
+            sponsorWallet,
+            deviationThreshold,
+            heartbeatInterval
+        );
+    }
+
+    function removeDapi(bytes32 dapiName) external {
```
bbenligiray, line 229: This needs to be behind the role check

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+        updateParameters = new UpdateParameters[](limitAdjusted);
+        for (uint256 ind = 0; ind < offset + limitAdjusted; ind++) {
+            bytes32 dapiNameHash = activeDapis.at(ind);
+            dapiNameHashes[ind] = dapiNameHash;
+            dataFeedIds[ind] = api3ServerV1.dapiNameHashToDataFeedId(
+                dapiNameHash
+            );
+            updateParameters[ind] = dapiToUpdateParameters[dapiNameHash];
+        }
+        // TODO: should this function also return the Signed API URLs for each Airnode in the UpdateParameters?
```
bbenligiray, line 267: Maybe. You need to be up to date with the airseeker-v2 discussions to be able to do this function correctly.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    function readDapis(
+        uint256 offset,
+        uint256 limit
+    )
+        external
+        view
+        returns (
+            bytes32[] memory dapiNameHashes,
```
bbenligiray, line 248: It feels like we need dAPI names, not their hashes

bbenligiray (commented):
Missing the registry update (because someone extended their subscription or upgraded the deviation threshold) but you probably know that

contracts/DapiDataRegistry.sol
acenolaza: I agree and I almost didn't add this parameter in the first place but I wasn't sure which values to use for the hash type. I guess I should just add the hash types as constants in here.

contracts/DapiDataRegistry.sol
acenolaza: I must admit I saw this comment coming 🤭 and I might have over-used the `struct` data structure (too many years of OOP might be the cause)

contracts/DapiDataRegistry.sol
acenolaza: This feels like it should require it's own role but at the same time I think the caller of this function is probably going to be the same as the caller of `registerDapi()` which is the `Registrar` role. WDYT?

A similar thing happens with `registerAirnodeSignedApiUrl()` function

contracts/DapiDataRegistry.sol
acenolaza: It felt more "correct" at the time since we are either sending a an encoded `bytes` for a `beacon` and encoded array of bytes for a `beaconSet`. The upside of this is that we don't have to worry about 2 arrays matching in length or in order of the elements but a downside is that it requires decoding in 2 steps (more gas). But I'll implement your suggestion since it probably makes more sense.

contracts/DapiDataRegistry.sol
bbenligiray: I've seen today that @metobom uses multiple OIS titles for an API provider because they have multiple base URLs, so this should definitely be left out because there is not a one-to-one mapping

contracts/DapiDataRegistry.sol
bbenligiray: Yeah, I think you can use the same role. I don't see your point about `registerAirnodeSignedApiUrl()` though. That's entirely MT-based/doesn't use roles?

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,267 @@
...
+            // https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/DataFeedServer.sol#L98
+            dataFeedId = keccak256(abi.encode(beaconIds));
+        }
+
+        dataFeedIdToData[dataFeedId] = dataFeedData;
+
+        emit RegisteredDataFeed(dataFeedId, dataFeedData);
+    }
+
+    function registerDapi(
```
bbenligiray, line 166: I'm a bit irked by `registerDapi()`/`removeDapi()` naming. They are functionally opposite but semantically remove is not the opposite of register.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,267 @@
...
+            bytes.concat(keccak256(abi.encode(airnode, url)))
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+
+        airnodeToSignedApiUrl[airnode] = url;
+
+        emit RegisteredSignedApiUrl(airnode, url);
+    }
+
+    function registerDatafeed(
```
bbenligiray, line 130:
```suggestion
    function registerDataFeed(
```

contracts/DapiDataRegistry.sol
acenolaza: I'm really glad you noticed that 🥇

contracts/DapiDataRegistry.sol
acenolaza: > I don't see your point about `registerAirnodeSignedApiUrl()` though. That's entirely MT-based/doesn't use roles?

You are absolutley right 🤦🏻 I got confused because Api3Market is supposed to call this function when making a purchase and Api3Market is going to need the role to be granted. I'll remove the access control for this function

contracts/DapiDataRegistry.sol
acenolaza: Options:
`registerDapi()`/`unregisterDapi()`
`addDapi()`/`removeDapi()`
`activateDapi()`/`deactivateDapi()`
`enableDapi()`/`disableDapi()`
`saveDapi()`/`deleteDapi()`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,267 @@
...
+    ) internal view returns (bool) {
+        return
+            manager == account ||
+            IAccessControlRegistry(accessControlRegistry).hasRole(
+                registrarRole,
+                account
+            );
+    }
+
+    function registerAirnodeSignedApiUrl(
```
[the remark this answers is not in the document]
acenolaza: > if it's not included in the current Merkle tree

I'm assuming this check is done off-chain and the function should be behind access control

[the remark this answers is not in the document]
alikonuk1: Essentially, instead of removing it, we should be updating it whenever we have a new root.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,312 @@
...
+        )
+    {
+        uint256 count = registeredDapisCount();
+        require(offset < count, "Invalid offset");
+        uint256 limitAdjusted = offset + limit > count ? count - offset : limit;
+        dapiNames = new bytes32[](limitAdjusted);
+        dataFeedIds = new bytes32[](limitAdjusted);
+        dataFeedDatas = new bytes[](limitAdjusted);
+        updateParameters = new UpdateParameters[](limitAdjusted);
+        for (uint256 ind = 0; ind < offset + limitAdjusted; ind++) {
```
alikonuk1, line 298: Shouldn't we start the ind at offset?

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,312 @@
...
+            dapiNames[ind] = dapiName;
+            bytes32 dapiNameHash = keccak256(abi.encodePacked(dapiName));
+            bytes32 dataFeedId = api3ServerV1.dapiNameHashToDataFeedId(
+                dapiNameHash
+            );
+            dataFeedIds[ind] = dataFeedId;
+            updateParameters[ind] = dapiNameHashToUpdateParameters[
+                dapiNameHash
+            ];
+            dataFeedDatas[ind] = dataFeedIdToData[dataFeedId];
```
alikonuk1, line 309: These would also go out of bound because of the comment above.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,300 @@
...
+    // string public constant PRICE_MANAGEMENT_HASH_TYPE_DESCRIPTION =
+    //     "Price management merkle tree root";
+    string public constant DAPI_MANAGEMENT_HASH_TYPE_DESCRIPTION =
+        "dAPI management merkle tree root";
+    string public constant API_INTEGRATION_HASH_TYPE_DESCRIPTION =
+        "API integration merkle tree root";
+
+    // bytes32 private constant _PRICE_MANAGEMENT_HASH_TYPE =
+    //     keccak256(abi.encodePacked(PRICE_MANAGEMENT_HASH_TYPE_DESCRIPTION));
+    bytes32 private constant _DAPI_MANAGEMENT_HASH_TYPE =
```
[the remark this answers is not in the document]
acenolaza: Is suggestion useful to make deployment cheaper? I used the same approach used for [roles and role descriptions](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/authorizers/RequesterAuthorizerWithAirnode.sol#L17-L21)

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,312 @@
...
+            hasRegistrarRoleOrIsManager(msg.sender),
+            "Sender is not manager or needs Registrar role"
+        );
+        require(airnode != address(0));
+        airnodeToSignedApiUrl[airnode] = "";
+
+        emit UnregisteredSignedApiUrl(airnode); // TODO: add msg.sender?
+    }
+
+    function registerDataFeed(
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/dapi-management/pull/3#discussion_r1360308361 I was wondering the same 🤭

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/bb514a149c05cc089260ad8c5fffad4f483a3d04

acenolaza:
I must admit I'm not convinced about https://github.com/api3dao/dapi-management/pull/3/commits/867c251ebe0f902605400dd2e4d283173cd6033e. I'm returning an array of Signed API URLs where it will contain a single element if `dataFeedData` is a `beacon` and multiple when it is a `beaconSet`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,339 @@
...
+        require(dataFeedData.length > 0, "Data feed data is empty");
+        bytes memory newDataFeedData;
+        if (dataFeedData.length == 64) {
+            // DataFeedId maps to a beacon
+            (address airnode, bytes32 templateId) = abi.decode(
+                dataFeedData,
+                (address, bytes32)
+            );
+
+            // TODO: check if signed API URL exists for Airnode?
```
acenolaza, line 170: This is something that came to mind after adding https://github.com/api3dao/dapi-management/pull/3/commits/867c251ebe0f902605400dd2e4d283173cd6033e

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    function unregisterAirnodeSignedApiUrl(address airnode) external {
+        require(
+            hasRegistrarRoleOrIsManager(msg.sender),
+            "Sender is not manager or needs Registrar role"
+        );
+        require(airnode != address(0));
+
+        // TODO: check if signed API URL is not being mappaed to an Airnode in dataFeedIdToData?
```
acenolaza, line 151: and this ...

contracts/DapiDataRegistry.sol
acenolaza: Added Signed API URLs in https://github.com/api3dao/dapi-management/pull/3/commits/867c251ebe0f902605400dd2e4d283173cd6033e

acenolaza:
I've added the Data Feed Data to the readDapis() function 

> I must admit I'm not convinced about [867c251](https://github.com/api3dao/dapi-management/commit/867c251ebe0f902605400dd2e4d283173cd6033e). I'm returning an array of Signed API URLs where it will contain a single element if `dataFeedData` is a `beacon` and multiple when it is a `beaconSet`

BTW even dataFeedData was not supposed to be included in the readDapis() function according to the slides 🤔 https://docs.google.com/document/d/1x5QBOGII8IUGjtoNR6PVE_UeqEjRQj2u3Ysa1FQkHf0/edit?disco=AAAA8P8fNmA

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,339 @@
...
+            bytes.concat(
+                keccak256(abi.encode(dapiName, dataFeedId, sponsorWallet))
+            )
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+
+        activeDapis.add(dapiName); // TODO: Not checking if already exists in set to allow for update parameters override
+        bytes32 dapiNameHash = keccak256(abi.encodePacked(dapiName));
+        dapiNameHashToUpdateParameters[dapiNameHash] = UpdateParameters(
+            deviationThresholdInPercentage, // TODO: can this be 0? should we check against any low/high boundary based on HUNDRED_PERCENT constant?
```
alikonuk1, line 246: I think we may want to keep our option to make these parameters 0 incase of a removal maybe. So not restricting it to be always bigger than 0 feels like a safer option for me.

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
alikonuk1: Saves 36973 gas on deployment, also feels tidier imo.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,338 @@
...
+
+    // This is updated using the API management merkle tree
+    mapping(address => string) public airnodeToSignedApiUrl;
+
+    // The value should be a single value or an array of them
+    // This needs to be encoded so we can determine if it's a beacon
+    // or a beaconSet based on the lenght
+    // It can be udpated by anyone because the contract will hash the data and derive it
+    // Airseeker will need to multicall to read all data using a single RPC call
+    mapping(bytes32 => bytes) public dataFeedIdToData;
```
Siegrift, line 75: (As of now) This is supposed to hold the template IDs of the data feed?

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
acenolaza: > Essentially, instead of removing it, we should be updating it whenever we have a new root.

I'm confused by this comment, you mind sharing an example or explain this a bit more 🙏🏻

contracts/DapiDataRegistry.sol
acenolaza: Yes, this is an encoded `(airnode, templateId)` or `(airnode[], templateId[])`

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
alikonuk1: So what I meant was, if we are registering `url` values to this `airnodeToSignedApiUrl[airnode]` mapping by validating the values with the API integration Merkle tree, wouldn't it be an antipattern to unregister them without validating the values with the Merkle tree?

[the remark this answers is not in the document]
acenolaza: Thanks, I get it now. I guess the issue is when you we stop supporting an API provider. How do you generate a proof for a value in the merkle tree which does not exist? I guess that is why I made the assumption this check was done off-chain and the remove function would be behind some access control role

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/5887df34c350861570f24679f77d2031cfe11d67

acenolaza:
I don't know if [this](https://github.com/api3dao/dapi-management/pull/3/commits/37590f61e5ab4f98728ef05d3105ff38b5d95ef0) is the right place for this change but I didn't want to forget about it. If a separate PR is preferable then I'll move it.

contracts/DapiDataRegistry.sol
bbenligiray: `addDapi()`/`removeDapi()` feels fine

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
bbenligiray: > How do you generate a proof for a value in the merkle tree which does not exist?

Huh, good point. I guess we have two options
- Let their signed API URLs remain in storage, it doesn't hurt us
- In the next Merkle tree, specify their signed API URL as `""` and register using that, which effectively deletes the previous one

Either way, we don't need a function to remove them. Feel free to resolve this thread.

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
bbenligiray: Revert this, I'd much rather pay the extra gas then take the risk of the hash going outdated

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+import "@api3/airnode-protocol-v1/contracts/access-control-registry/AccessControlRegistryAdminnedWithManager.sol";
+import "@api3/airnode-protocol-v1/contracts/api3-server-v1/interfaces/IApi3ServerV1.sol";
+import "@api3/airnode-protocol-v1/contracts/utils/SelfMulticall.sol";
+import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import "./interfaces/IDapiDataRegistry.sol";
+import "./interfaces/IHashRegistry.sol";
+
+contract DapiDataRegistry is
+    SelfMulticall,
```
bbenligiray, line 13: This is redundant, `AccessControlRegistryAdminnedWithManager` inherits `SelfMulticall`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    AccessControlRegistryAdminnedWithManager,
+    IDapiDataRegistry
+{
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    /// @notice Number that represents 100%
+    uint256 public constant override HUNDRED_PERCENT = 1e8;
+
+    // dAPI management merkle tree root
+    bytes32 public constant override DAPI_MANAGEMENT_HASH_TYPE =
```
bbenligiray, line 23: Do this in a single line as in https://github.com/api3dao/omnimarket/blob/main/contracts/Omnimarket.sol#L81

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        0xf607eab1a0e1d5843e97cc3768147f5f15420188755ebc3a685ddcefe6d79d63;
+
+    /// @notice Registrar role description
+    string public constant override REGISTRAR_ROLE_DESCRIPTION = "Registrar";
+
+    /// @notice Registrar role
+    bytes32 public immutable override registrarRole;
+
+    IHashRegistry public immutable override hashRegistry;
+    IApi3ServerV1 public immutable override api3ServerV1;
```
bbenligiray, line 36: I declare these as `address` to avoid having to import `IHashRegistry.sol` and `IApi3ServerV1.sol` to `IDapiDataRegistry.sol`. I'm not sure if calling `IApi3ServerV1()` and `IDapiDataRegistry()` at runtime has an overhead though.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    SelfMulticall,
+    AccessControlRegistryAdminnedWithManager,
+    IDapiDataRegistry
+{
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    /// @notice Number that represents 100%
+    uint256 public constant override HUNDRED_PERCENT = 1e8;
+
+    // dAPI management merkle tree root
```
bbenligiray, line 22: Capitalize Merkle

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    // This is updated using the API management merkle tree
+    // TODO: should this mapping be private now that we are returning these valuse via readDapis()
+    mapping(address => string) public override airnodeToSignedApiUrl;
+
+    // The value should be a single value or an array of them
+    // This needs to be encoded so we can determine if it's a beacon
+    // or a beaconSet based on the lenght
+    // It can be udpated by anyone because the contract will hash the data and derive it
+    // Airseeker will need to multicall to read all data using a single RPC call
+    // TODO: should this mapping be private now that we are returning these valuse via readDapis()
```
bbenligiray, line 47: No

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    mapping(bytes32 => bytes) public override dataFeedIdToData;
```
bbenligiray, line 48: Maybe call this `dataFeeds`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+
+    // This is the list of dAPIs AirseekerV2 will need to update
+    // Api3Market contract will have a role to update this after a purchase
+    // Dapi names are expected to be unique bytes32 strings
+    EnumerableSet.Bytes32Set private activeDapis;
```
bbenligiray, line 53: Maybe call this `dapis` (since there is no `passiveDapis`)

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+
+    mapping(bytes32 => UpdateParameters) private dapiNameHashToUpdateParameters;
```
bbenligiray, line 55: Make this `dapiNameToUpdateParameters`. Also it shouldn't be private.

contracts/DapiDataRegistry.sol
bbenligiray: Also shouldn't be private

contracts/DapiDataRegistry.sol
bbenligiray: btw these can be private

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+
+    constructor(
+        address _accessControlRegistry,
+        string memory _adminRoleDescription,
+        address _manager,
+        IHashRegistry _hashRegistry,
+        IApi3ServerV1 _api3ServerV1
```
bbenligiray, line 62: Validate that these are not zero

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            REGISTRAR_ROLE_DESCRIPTION
+        );
+        hashRegistry = _hashRegistry;
+        api3ServerV1 = _api3ServerV1;
+    }
+
+    /// @dev Returns if the account has the Registrar role or is the manager
+    /// @param account Account address
+    /// @return If the account has the Registrar role or is the manager
+    function hasRegistrarRoleOrIsManager(
```
bbenligiray, line 81: You're only using this internally with `msg.sender`, so I don't see why this is not a modifier

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            bytes.concat(keccak256(abi.encode(airnode, url)))
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+
+        airnodeToSignedApiUrl[airnode] = url;
+
+        emit RegisteredSignedApiUrl(airnode, url);
+    }
+
+    function unregisterAirnodeSignedApiUrl(address airnode) external override {
```
bbenligiray, line 117: Remove this

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            // Derive beacon ID
+            // https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/DataFeedServer.sol#L87
+            dataFeedId = keccak256(abi.encodePacked(airnode, templateId));
+            newDataFeedData = dataFeedData;
+        } else {
+            // dataFeedData must have an even number of bytes32 pairs
+            require(
+                (dataFeedData.length / 2) % 32 == 0,
+                "Invalid data feed data"
+            );
```
bbenligiray, line 154: A `dataFeedData.length` of 65 will not revert here. This is a much safer way of checking
```
(address[] memory airnodes, bytes32[] memory templateIds) = abi
                .decode(dataFeedData, (address[], bytes32[]));
require(abi.encode(airnode, templateIds).length == dataFeedData.length);
```

bbenligiray: This may also revert if `dataFeedData.length` is 0 so the validation above becomes redundant, gotta check

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        delete airnodeToSignedApiUrl[airnode];
+
+        emit UnregisteredSignedApiUrl(airnode); // TODO: add msg.sender?
+    }
+
+    function registerDataFeed(
+        bytes calldata dataFeedData
+    ) external override returns (bytes32 dataFeedId) {
+        require(dataFeedData.length > 0, "Data feed data is empty");
+        bytes memory newDataFeedData;
```
bbenligiray, line 135: You shouldn't need this. Just revert if `dataFeedData` has trailing bytes and store `dataFeedData` otherwise.

contracts/DapiDataRegistry.sol
bbenligiray: No

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        bytes32 dataFeedId,
+        address sponsorWallet,
+        uint256 deviationThresholdInPercentage,
+        int224 deviationReference,
+        uint32 heartbeatInterval,
+        bytes32 root,
+        bytes32[] calldata proof
+    ) external override {
+        require(root != bytes32(0), "Root is zero");
+        require(proof.length != 0, "Proof is empty");
```
bbenligiray, line 192: No need to validate these. You are validating `dapiName` in the function below though, so validate the others consistently.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+
+        // Verify proof
+        bytes32 leaf = keccak256(
+            bytes.concat(
+                keccak256(abi.encode(dapiName, dataFeedId, sponsorWallet))
+            )
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+
+        activeDapis.add(dapiName); // TODO: Not checking if already exists in set to allow for update parameters override (downgrade/upgrade)
```
bbenligiray, line 216: Okay. Mention in `@dev` that the registrar is responsible with validating how the parameters are being overwritten

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        bytes32 dapiNameHash = keccak256(abi.encodePacked(dapiName));
+        dapiNameHashToUpdateParameters[dapiNameHash] = UpdateParameters(
+            deviationThresholdInPercentage, // TODO: can this be 0? should we check against any low/high boundary based on HUNDRED_PERCENT constant?
```
bbenligiray, line 219: I think so? Gotta check what Airseeker does when it's configured as 0. Validating this is Api3Market's responsibility.

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            deviationReference,
+            heartbeatInterval // TODO: can this be 0?
```
bbenligiray, line 221: Same as above

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        require(dapiName != bytes32(0), "dAPI name is zero");
+        require(
+            hasRegistrarRoleOrIsManager(msg.sender),
+            "Sender is not manager or needs Registrar role"
+        );
+        require(activeDapis.remove(dapiName), "dAPI name is not registered");
+        bytes32 dapiNameHash = keccak256(abi.encodePacked(dapiName));
+        delete dapiNameHashToUpdateParameters[dapiNameHash];
+
+        emit UnregisteredDapi(dapiName); // TODO: add msg.sender?
```
bbenligiray, line 247: Yes

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        returns (
+            bytes32[] memory dapiNames,
+            bytes32[] memory dataFeedIds,
+            UpdateParameters[] memory updateParameters,
+            bytes[] memory dataFeedDatas,
+            string[][] memory signedApiUrls
+        )
+    {
+        uint256 count = registeredDapisCount();
+        require(offset < count, "Invalid offset");
```
bbenligiray, line 275: Return empty instead

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    function readDapis(
+        uint256 offset,
+        uint256 limit
+    )
+        external
+        view
+        override
+        returns (
+            bytes32[] memory dapiNames,
+            bytes32[] memory dataFeedIds,
```
bbenligiray, line 268: This is redundant, can be derived from `dataFeedDatas`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            UpdateParameters[] memory updateParameters,
+            bytes[] memory dataFeedDatas,
```
bbenligiray, line 270: Call this `dataFeeds`

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    function registeredDapisCount()
+        public
+        view
+        override
+        returns (uint256 count)
+    {
+        count = activeDapis.length();
+    }
+
+    function readDapis(
```
bbenligiray, line 259: This function should also call `api3ServerV1.dataFeeds(dataFeedId)` for each dAPI and return them

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+        require(offset < count, "Invalid offset");
+        uint256 limitAdjusted = offset + limit > count ? count - offset : limit;
+        dapiNames = new bytes32[](limitAdjusted);
+        dataFeedIds = new bytes32[](limitAdjusted);
+        updateParameters = new UpdateParameters[](limitAdjusted);
+        dataFeedDatas = new bytes[](limitAdjusted);
+        signedApiUrls = new string[][](limitAdjusted);
+        for (uint256 i = offset; i < offset + limitAdjusted; i++) {
+            bytes32 dapiName = activeDapis.at(i);
+            uint256 currentIndex = i - offset;
```
bbenligiray, line 284: `i` is short for index so having an `i` and a `currentIndex` in the same scope is confusing

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+            bytes memory dataFeedData = dataFeedIdToData[dataFeedId];
+            dataFeedDatas[currentIndex] = dataFeedData;
+            if (dataFeedData.length == 64) {
+                (address airnode, ) = abi.decode(
+                    dataFeedData,
+                    (address, bytes32)
+                );
+                string[] memory urls = new string[](1);
+                urls[0] = airnodeToSignedApiUrl[airnode];
+                signedApiUrls[currentIndex] = urls;
```
bbenligiray, line 303: Something like this would be much smaller
```
signedApiUrls = abi.encodePacked(signedApiUrls, "\x1F", airnodeToSignedApiUrl[airnode]);
```

bbenligiray: btw I looked at the returned data and returning a signed API URL per Beacon doesn't look that bad

bbenligiray (changes requested):
Let's merge this PR asap

contracts/DapiDataRegistry.sol
acenolaza: I don't think Bytes32Set can be public. I get this when I try:

`Internal or recursive type is not allowed for public state variables.solidity(6744)`

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/22b984646929683b302d971cc7dcfb1f2a511f71

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/19c4d4121797be51c4b5a97764b47981c982cd1f

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/5e3e7b4989294df47ce7dff1ba6991c26f0b23bd

contracts/DapiDataRegistry.sol
acenolaza: https://github.com/api3dao/dapi-management/pull/3/commits/507922711d2360839b735f5de81de50d574920cd

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/8f94c2eeb2318e0acbee1e5af30f393fba72655b

contracts/DapiDataRegistry.sol
acenolaza: https://github.com/api3dao/dapi-management/pull/3/commits/26af162c80e73e85b83711f6d7d1397dc42d45d1

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    string public constant override REGISTRAR_ROLE_DESCRIPTION = "Registrar";
+
+    /// @notice Registrar role
+    bytes32 public immutable override registrarRole;
+
+    IHashRegistry public immutable override hashRegistry;
+    IApi3ServerV1 public immutable override api3ServerV1;
+
+    // This is updated using the API management merkle tree
+    // TODO: should this mapping be private now that we are returning these valuse via readDapis()
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/dapi-management/pull/3/commits/5bd338872c73b728feaf046d8b2a537e41e6b7b0

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/5bd338872c73b728feaf046d8b2a537e41e6b7b0

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/1401563ad410cbdb2825d749c96a35d5e8022c42

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/44cf3620f2b3cc3dabca1d2795d4ef73fc7d70cf

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/8f94c2eeb2318e0acbee1e5af30f393fba72655b

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/76d5e37381e1ea5883f4b8ec722721f8b92e2eff

contracts/DapiDataRegistry.sol
acenolaza: Had to call it `dataFeeds_` to avoid naming collision with state var

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/6d5f5c5d75dde09d8acee4f27dd52573679aacb1

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    function registerAirnodeSignedApiUrl(
+        address airnode,
+        string calldata url,
+        bytes32 root,
+        bytes32[] calldata proof
+    ) external override {
+        require(root != bytes32(0), "Root is zero");
+        require(proof.length != 0, "Proof is empty");
```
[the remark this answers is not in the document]
acenolaza: I thought this was a gas optimization thing and that is why it was also present [here](https://github.com/api3dao/manager-multisig/blob/main/chain/contracts/DapiFallbackV1.sol#L461-L466). I apply the suggested changes anyways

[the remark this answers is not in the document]
acenolaza: @bbenligiray  actually, are you sure about this? If `airnode` is zero then the `root` will need to be something different from what was registered in the `HashRegistry`.  One thing I do wonder now that I'm looking at this is if we should check that the `root.timestamp` is not older than X days or something and not only check that the `root` was registered on the `HashRegistry`.

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/5b3dd072be669df095fe6122d80de02740013d02

contracts/DapiDataRegistry.sol
acenolaza: I'm not too sure about this but I think it's fixed by https://github.com/api3dao/dapi-management/pull/3/commits/72c55bc1301ffb3f0f4e51bf8808dbefafa4e3c7

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,317 @@
...
+                (dataFeedData.length / 2) % 32 == 0,
+                "Invalid data feed data"
+            );
+            // DataFeedId maps to a beaconSet
+            (address[] memory airnodes, bytes32[] memory templateIds) = abi
+                .decode(dataFeedData, (address[], bytes32[]));
+            require(airnodes.length == templateIds.length, "Length mismatch");
+            bytes32[] memory beaconIds = new bytes32[](airnodes.length);
+            for (uint256 i = 0; i < airnodes.length; i++) {
+                // TODO: check if signed API URL exists for Airnode?
```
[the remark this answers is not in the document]
acenolaza: https://github.com/api3dao/dapi-management/pull/3/commits/8ab2ed2355c5686f005765a8b2f1c36d5dcd971a

contracts/DapiDataRegistry.sol
acenolaza: Waiting for [this question](https://github.com/api3dao/dapi-management/pull/3#discussion_r1372137065) to be answered...

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/03b7f4cc3f29ddbb35fe879fd079e84d59a46a36

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/a9ad9ea6c6f9bf9b8ba357424ce2331bdbe6d585

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/04059fe3904bbefd842fcbca407b8bfcfa494824

contracts/DapiDataRegistry.sol
acenolaza: Maybe then: `ind`, `currentInd` and `ind2` https://github.com/api3dao/dapi-management/pull/3/commits/c1e63f96a94443e83de804662316a24249065bcc

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/b01c277708e7a56c31cbccd44637865fc589218c

contracts/DapiDataRegistry.sol
acenolaza: Currently it seems that deviation = 0 is equal to "update always" in Airseeker:
1. not validating value from config: https://github.com/api3dao/airseeker/blob/main/src/validation.ts#L137
2. not checking in condition function: https://github.com/api3dao/airseeker/blob/main/src/check-condition.ts#L44

contracts/DapiDataRegistry.sol
acenolaza: and how do you feel about packing each Airnode Signed API URL along with each data feed data encoded object? This doesn't dedup the urls and requires the use of standard `abi.encode` but it makes the payload simpler to use:

```
    function readDapis(
        uint256 offset,
        uint256 limit
    )
        external
        view
        override
        returns (
            bytes32[] memory dapiNames,
            UpdateParameters[] memory updateParameters,
            DataFeedValues[] memory dataFeedValues,
            bytes[] memory dataFeedsWithSignedApiUrls
        )
    {
        uint256 count = dapisCount();
        if (offset < count) {
            uint256 limitAdjusted = offset + limit > count
                ? count - offset
                : limit;
            dapiNames = new bytes32[](limitAdjusted);
            updateParameters = new UpdateParameters[](limitAdjusted);
            dataFeedValues = new DataFeedValues[](limitAdjusted);
            dataFeedsWithSignedApiUrls = new bytes[](limitAdjusted);
            for (uint256 ind = offset; ind < offset + limitAdjusted; ind++) {
                bytes32 dapiName = _dapis.at(ind);
                uint256 currentInd = ind - offset;
                dapiNames[currentInd] = dapiName;
                bytes32 dapiNameHash = keccak256(abi.encodePacked(dapiName));
                bytes32 dataFeedId = IApi3ServerV1(api3ServerV1)
                    .dapiNameHashToDataFeedId(dapiNameHash);
                (int224 value, uint32 timestamp) = IApi3ServerV1(api3ServerV1)
                    .dataFeeds(dataFeedId);
                dataFeedValues[currentInd] = DataFeedValues(value, timestamp);
                updateParameters[currentInd] = dapiNameToUpdateParameters[
                    dapiNameHash
                ];
                bytes memory dataFeed = dataFeeds[dataFeedId];
                if (dataFeed.length == 64) {
                    (address airnode, bytes32 templateId) = abi.decode(
                        dataFeed,
                        (address, bytes32)
                    );
                    dataFeedsWithSignedApiUrls[currentInd] = abi.encode(
                        airnode,
                        templateId,
                        airnodeToSignedApiUrl[airnode]
                    );
                } else {
                    (
                        address[] memory airnodes,
                        bytes32[] memory templateIds
                    ) = abi.decode(dataFeed, (address[], bytes32[]));
                    string[] memory urls = new string[](airnodes.length);
                    for (uint256 ind2 = 0; ind2 < airnodes.length; ind2++) {
                        urls[ind2] = airnodeToSignedApiUrl[airnodes[ind2]];
                    }
                    dataFeedsWithSignedApiUrls[currentInd] = abi.encode(
                        airnodes,
                        templateIds,
                        urls
                    );
                }
            }
        }
    }
```

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/49eae4e5929a4211546c12cc056bacfaff54fc6f

contracts/DapiDataRegistry.sol
acenolaza: I know even think that the check on `dapiName` is redundant on `removeDapi()` below so I removed it https://github.com/api3dao/dapi-management/pull/3/commits/4e8e5c5cddbec5a8d41d4452f22d68e2fcf17a63

contracts/DapiDataRegistry.sol
acenolaza: In `HashRegistry` I workaround it by doing [this](https://github.com/api3dao/dapi-management/blob/169-DapiDataRegistry/contracts/HashRegistry.sol#L107C9-L107C9)

contracts/DapiDataRegistry.sol
bbenligiray: I'm on the edge but leaning towards not liking it because a signed API URL belongs to an Airnode and not a data feed, associating it with the data feed comes off as wrong to me (though the alternative also does that implicitly).

> it makes the payload simpler to use

I'm not seeing that tbh. You already need to match dAPI names to data feeds, also requiring to match the signed API URLs to data feeds doesn't feel too much at that point. What would actually be easy to use for this to return a `Dapi` struct that includes both the name, data feeds, their URLs, etc.

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
bbenligiray: > If airnode is zero then the root will need to be something different from what was registered in the HashRegistry

Not if the Merkle tree whose root is registered at `HashRegistry` includes a zero `airnode` address

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+    mapping(bytes32 => bytes) public override dataFeeds;
+
+    /// @notice Parameters used while checking the conditions for updating a dAPI
+    mapping(bytes32 => UpdateParameters)
+        public
+        override dapiNameToUpdateParameters;
+
+    EnumerableSet.Bytes32Set private _dapis;
+
+    bytes32 private constant _DAPI_MANAGEMENT_HASH_TYPE =
```
bbenligiray, line 65:
```suggestion
    bytes32 private constant DAPI_MANAGEMENT_MERKLE_TREE_ROOT_HASH_TYPE =
```

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+        keccak256(abi.encodePacked("dAPI management Merkle tree root"));
+    bytes32 private constant _SIGNED_API_URL_HASH_TYPE =
+        keccak256(abi.encodePacked("Signed API URL Merkle root"));
```
bbenligiray, line 68:
```suggestion
    bytes32 private constant SIGNED_API_URL_MERKLE_TREE_ROOT_HASH_TYPE =
        keccak256(abi.encodePacked("Signed API URL Merkle tree root"));
```

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+    /// @dev The length of the bytes is used to determine if the encoded data
+    /// feed data belongs to a beacon or beacon set
+    mapping(bytes32 => bytes) public override dataFeeds;
+
+    /// @notice Parameters used while checking the conditions for updating a dAPI
+    mapping(bytes32 => UpdateParameters)
+        public
+        override dapiNameToUpdateParameters;
+
+    EnumerableSet.Bytes32Set private _dapis;
```
bbenligiray, line 63:
```suggestion
    EnumerableSet.Bytes32Set private dapis;
```

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+        keccak256(abi.encodePacked("Signed API URL Merkle root"));
+
+    modifier onlyRegistrarOrManager() {
+        require(
+            msg.sender == manager ||
+                IAccessControlRegistry(accessControlRegistry).hasRole(
+                    registrarRole,
+                    msg.sender
+                ),
+            "Sender is not manager or needs Registrar role"
```
bbenligiray, line 77:
```suggestion
            "Sender is not manager or has Registrar role"
```

contracts/DapiDataRegistry.sol
bbenligiray: Imo Merkle tree values need to be validated wherever possible https://github.com/api3dao/dapi-management/pull/17#discussion_r1374555280

contracts/DapiDataRegistry.sol
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    /// @notice Called to register data about a data feed
+    /// @dev Data feed IDs are derived based on this data. If the encoded bytes
+    /// has a length of 64 it is considered to be data for deriving a beacon ID.
+    /// Otherwise it is considered data for deriving beacon IDs that are then
+    /// used to derive the beacon set ID
+    /// @param dataFeed Encoded data feed data
+    function registerDataFeed(
+        bytes calldata dataFeed
```
bbenligiray, line 150: Check if `else` reverts when `dataFeed.length == 0` and revert explicitly if not

contracts/DapiDataRegistry.sol
acenolaza: I guess this becomes obsolete now that [readDapis()](https://github.com/api3dao/dapi-management/pull/3/commits/4aceef94c1911eaccd7b71bf2d786c313830224f) no longer exists and dAPIs must be fetched one by one (via multicall) based on [this](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1698326968765089) and [this](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1698327049106809) comments.

I don't have a strong opinion on using or not a `Dapi` struct but one thing that is now a concern for me is: how difficult would be it be for callers to decode the existing structs (`UpdateParameters` and `DataFeedValue`) with something like ethers.js when calling this functions via static `multicall()` which returns [encoded](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/utils/SelfMulticall.sol#L21) results.

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/2682e771c539eebf2e13cc21d893b3d0f689d840

contracts/DapiDataRegistry.sol
acenolaza: Good catch. Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/2682e771c539eebf2e13cc21d893b3d0f689d840

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/a088d2459cf1674f69c6de5b4fc29a2cb8ff01ba

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/e7339c885fe1cfe90ccaf2eb51105875b7379262

contracts/DapiDataRegistry.sol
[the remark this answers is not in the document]
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/05d1ed6880dfe2bfc14159f4b1b613ff3e34029a

contracts/DapiDataRegistry.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/b0f812b515ef0563f4bf8ad7fff2df83535d66cc

contracts/DapiDataRegistry.sol
acenolaza: I have a[ test for this case](https://github.com/api3dao/dapi-management/blob/169-DapiDataRegistry/test/DapiDataRegistry.sol.js#L437) and it does revert without reason so I'll add a `require()`

acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/3/commits/00b25a6b52d93382664fa04fc955fe53ba12065d

PR #17, opened 2023-10-24
WIP: Adds DapiFallbackV2.sol contract

alikonuk1:


package.json
```diff
@@ -21,6 +21,7 @@
     "prettier": "prettier --write \"./**/*.{js,ts,md,json,sol}\"",
     "test": "hardhat test",
     "test:coverage": "hardhat coverage",
+    "test:gas": "REPORT_GAS=TRUE hardhat test",
```
alikonuk1, line 24: Not sure but small enough (and necessary) change that can be also added with this pr?

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+    bytes32 private constant _PRICE_HASH_TYPE =
+        keccak256(abi.encodePacked("Price Merkle tree root"));
+
+    constructor(address _api3ServerV1, address _hashRegistry) {
+        require(
+            _api3ServerV1 != address(0) && _hashRegistry != address(0),
+            "Address cannot be zero"
+        );
+        api3ServerV1 = _api3ServerV1;
+        hashRegistry = _hashRegistry;
```
bbenligiray, line 25: Validate that these are not zero

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    function withdraw(
+        address payable recipient,
+        uint256 amount
+    ) external override onlyOwner {
+        _withdraw(recipient, amount);
+    }
+
+    function executeDapiFallback(
+        ExecuteDapiFallbackArgs calldata args
```
bbenligiray, line 38: Can we not unroll these?

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+    ) external override {
+        bytes32 currentBeaconId = IApi3ServerV1(api3ServerV1)
```
bbenligiray, line 40: This is not necessarily a Beacon

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+            .dapiNameHashToDataFeedId(args.dapiName);
+        require(
+            currentBeaconId != args.beaconId,
```
bbenligiray, line 43: The fallback data feed is also not necessarily a Beacon

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+        );
+
+        _validateTree(
+            _PRICE_HASH_TYPE,
+            args.priceProof,
+            args.priceRoot,
+            priceLeaf
+        );
+
+        uint256 minSponsorWalletBalance = (args.price * 86400) / args.duration;
```
bbenligiray, line 82:
```suggestion
        uint256 minSponsorWalletBalance = (args.price * 1 days) / args.duration;
```

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    function _withdraw(address payable recipient, uint256 amount) private {
+        require(recipient != address(0), "Recipient address zero");
+        require(amount != 0, "Amount zero");
+        require(
+            address(this).balance >= amount,
+            "Insufficient contract balance"
+        );
+        (bool success, ) = recipient.call{value: amount}("");
+        require(success, "Failed to withdraw");
```
bbenligiray, line 116: Use Address.sol

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+        bytes32 leaf
+    ) private view {
+        require(
+            IHashRegistry(hashRegistry).hashTypeToHash(treeType) == root,
+            "Tree has not been registered"
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid tree proof");
+    }
+
+    function _withdraw(address payable recipient, uint256 amount) private {
```
bbenligiray, line 108: I don't see the point in having a `_withdraw()` if it's not being called in multiple places. Functions calls are an overhead so move this into `withdraw()`.

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+            "Address cannot be zero"
+        );
+        api3ServerV1 = _api3ServerV1;
+        hashRegistry = _hashRegistry;
+    }
+
+    receive() external payable {}
+
+    function withdraw(
+        address payable recipient,
```
bbenligiray, line 31: I'm still not sure if a `withdraw()` function should have a `recipient` argument. I don't have it in OevAuctionHouse.sol.

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(
+            address(this).balance >= amount,
+            "Insufficient contract balance"
+        );
+        (bool success, ) = recipient.call{value: amount}("");
+        require(success, "Failed to withdraw");
+        emit Withdrawn(recipient, amount, address(this).balance);
+    }
+
+    function _fundSponsorWallet(
```
bbenligiray, line 120: Same thing, you can embed this in `executeDapiFallback()`

hardhat.config.js
```diff
@@ -6,12 +6,19 @@ module.exports = {
   etherscan: hardhatConfig.etherscan(),
   networks: hardhatConfig.networks(),
   solidity: {
-    version: '0.8.18',
-    settings: {
-      optimizer: {
-        enabled: true,
-        runs: 200,
+    compilers: [
```
bbenligiray, line 9: I was too lazy to do this so I bumped the OevAuctionHouse.sol version down to 0.8.17 :see_no_evil:

bbenligiray (commented):
I can do another pass once these are resolved but already looks pretty solid

contracts/DapiFallbackV2.sol
bbenligiray: I guess be 100% sure that changing the order operations in the function doesn't solve the problem before resorting to this

alikonuk1: Tried different things before going with this option but none of them worked.

BTT21000:
Should the [Natspec](https://docs.soliditylang.org/en/latest/natspec-format.html) comments be typically added to the contract code as it progresses (see example of comments on the [HashRegistry](https://github.com/api3dao/dapi-management/blob/109-dapifallbackv2/contracts/HashRegistry.sol) contract) ?

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        _validateTree(
+            _PRICE_HASH_TYPE,
+            args.priceProof,
+            args.priceRoot,
+            priceLeaf
+        );
+
+        uint256 minSponsorWalletBalance = (args.price * 1 days) / args.duration;
```
acenolaza, line 90: Should we check that `args.duration` is not zero or do we just trust the merkle tree to validate this?

alikonuk1: Yes, so we would need to check it if we have this line before the tree validation. But validation from _validateTree() should be enough here.

contracts/DapiFallbackV2.sol
bbenligiray: Okay

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+/// @dev The contract inherits from the Ownable contract of the OpenZeppelin library
+/// @dev which provides basic authorization control functions.
+contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
+    /// @notice Api3ServerV1 contract address
+    address public immutable api3ServerV1;
+    /// @notice HashRegistry contract address
+    address public immutable hashRegistry;
+
+    /// @notice Constants defining types of the Merkle tree roots used within the contract logic.
+    bytes32 private constant _DAPI_FALLBACK_HASH_TYPE =
```
bbenligiray, line 23:
```suggestion
```

bbenligiray, line 23:
```suggestion
    bytes32 private constant DAPI_FALLBACK_MERKLE_TREE_ROOT_HASH_TYPE =
```

bbenligiray: Afaik there is nothing in the official style guide about private variables having a leading underscore

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+        keccak256(abi.encodePacked("dAPI fallback Merkle tree root"));
+    bytes32 private constant _PRICE_HASH_TYPE =
```
bbenligiray, line 25:
```suggestion
    bytes32 private constant DAPI_PRICING_MERKLE_TREE_ROOT_HASH_TYPE =
        keccak256(abi.encodePacked("dAPI pricing Merkle tree root"));
```

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    /// @notice Initializes the contract setting the api3ServerV1 and hashRegistry addresses.
+    /// @param _api3ServerV1 The address of the Api3ServerV1 contract.
+    /// @param _hashRegistry The address of the HashRegistry contract.
+    /// @dev The constructor requires non-zero addresses for the api3ServerV1 and hashRegistry contracts.
+    constructor(address _api3ServerV1, address _hashRegistry) {
+        require(
+            _api3ServerV1 != address(0) && _hashRegistry != address(0),
+            "Address cannot be zero"
+        );
```
bbenligiray, line 36: We do this kind of stuff separately everywhere else

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+    receive() external payable {}
+
+    /// @notice Allows the contract owner to withdraw funds from the contract.
+    /// @param amount The amount of funds to withdraw.
+    function withdraw(uint256 amount) external override onlyOwner {
+        require(amount != 0, "Amount zero");
+        require(
+            address(this).balance >= amount,
+            "Insufficient contract balance"
+        );
```
bbenligiray, line 52: This is redundant, Address.sol checks it

contracts/DapiFallbackV2.sol
bbenligiray: Auditors will 100% flag this so validate that it's not zero

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        uint256 minSponsorWalletBalance = (args.price * 1 days) / args.duration;
+
+        uint256 sponsorWalletBalance = args.sponsorWallet.balance;
+        if (sponsorWalletBalance < minSponsorWalletBalance) {
+            uint256 amount = minSponsorWalletBalance - sponsorWalletBalance;
+            require(
+                address(this).balance >= amount,
+                "Insufficient contract balance"
+            );
```
bbenligiray, line 133: Redundant

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+            Address.sendValue(args.sponsorWallet, amount);
```
bbenligiray, line 134: `args.sponsorWallet` is trusted less than `api3ServerV1` in the context of reentrancy. Do this call the last.

contracts/DapiFallbackV2.sol
```diff
@@ -0,0 +1,163 @@
...
+    ///   - `fallbackProof`: Merkle proof for validating the dAPI fallback parameters.
+    ///   - `updateParams`: Encoded parameters for the data update.
+    ///   - `priceRoot`: Root of the Merkle tree for the pricing data.
+    ///   - `priceProof`: Merkle proof for validating the pricing data.
+    ///   - `duration`: Time period for which the price is calculated.
+    ///   - `price`: Cost of the data feed for a given duration.
+    ///   - `sponsorWallet`: Address of the sponsor wallet,
+    /// which is funded if the balance is below the required minimum.
+    function executeDapiFallback(
+        ExecuteDapiFallbackArgs calldata args
```
bbenligiray, line 77: I think these should be validated even though they are proven by the tree because the integrity of the tree is proven by signatures after all, it is not trustless

bbenligiray (approved):
You can merge after addressing the comments

PR #18, opened 2023-10-25
Add ability to connect a wallet

### LINKED ISSUE #15: Add ability to connect a wallet
We don't need to go all out and use Wallet Connect. Connecting to MetaMask should be good enough.

mcoetzee:
Closes #15 

### What does this change?
- adds an account section in die side panel with
  - Connect button when disconnected
  - the user's address when connected (the wallet will stay connected until it is disconnected via MetaMask)
  - a popover that informs the user that they can disconnect via MetaMask (it is not possible to programatically disconnect)
- provides a `useWeb3Data()` hook that provides you with the `provider`, `signer`, `address`, `chainId`, `connectStatus` etc
- adds the `Buttton` and `Popover` components with https://ui.shadcn.com/
- adds a script to use when adding new shadcn components: `yarn frontend:add-component x`

### Screenshots

https://github.com/api3dao/dapi-management/assets/747979/5b548909-73ae-430e-a09a-8b16e1883765

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,9 +9,4 @@ module.exports = {
     sourceType: 'module',
   },
   extends: 'eslint:recommended',
-  settings: {
-    next: {
-      rootDir: 'frontend/',
-    },
-  },
 };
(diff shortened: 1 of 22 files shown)


BTT21000:
It looks great! Just a minor UX comment: I am wondering if the connect button should be at the top right of the page like standard DApps ?

pieterdub (approved):
LGTM, chadcn is great 👏

PR #26, opened 2023-10-27
Api3Market contract #170

acenolaza:
~** THIS IS A WIP **~

Created a ginormous `buyDapi()` function because I wanted to lay out all the initial requirements. We can split this function and ask the caller to use `multicall` but we might need to add checks to functions that require one to be call before the other.

As usual, there a bunch of TODOs with questions and early feedback is always welcome 🙏🏻


contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+    function buyDapi(BuyDapiArgs calldata args) external payable override {
+        // Store Signed API URLs for all the Airnodes used by the constituent beacons of the beaconSet
+        require(
+            args.airnodes.length == args.urls.length &&
+                args.urls.length == args.signedApiUrlProofs.length,
+            "Airondes, URLs or Signed API URL proofs length mismatch"
+        );
+        for (uint ind = 0; ind < args.signedApiUrlProofs.length; ind++) {
+            // TODO: The being called is very naive and does not check if url being registered is the same for the current airnode
+            //       Should we add that check to avoid re-setting the same value to state if values are equal?
```
bbenligiray, line 60: Multi-call DapiDataRegistry to fetch all signed API URLs, only register the ones that don't match

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+                args.signedApiUrlProofs[ind]
+            );
+        }
+
+        // Store the actual data used to derive each beaconId that will then be used to derive the beaconSetId
+        require(
+            args.airnodes.length == args.templateIds.length,
+            "Airnodes and template IDs length mismatch"
+        );
+        // TODO: dAPI purchases will always use beaconSets or will there be purchases that will point to a single beacon?
```
bbenligiray, line 74: There will be single Beacon dAPIs

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        IDapiDataRegistry(dapiDataRegistry).registerDataFeed(
+            abi.encode(args.airnodes, args.templateIds)
+        );
+
+        // Store the dAPI name along with the update parameters used by AirseekerV2
+        // TODO: handle downgrade/upgrade
+        //       say we have 0.25% active for the next 3 months and someone wants
+        //       to come in and buy 1% for the next 6 months, which means they
+        //       should only pay for 1% for 3 months and the dAPI to be
+        //       downgraded to 1% after 3 months
```
bbenligiray, line 84: Yes, ideally

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        // TODO: should we derive the data feed ID based on the airnodes and templateIds to see if it matches the dataFeedId parameter?
```
bbenligiray, line 85: Why would there be a dataFeedId parameter to begin with?

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+                    )
+                )
+            )
+        );
+        require(
+            MerkleProof.verify(args.priceProof, args.priceRoot, priceLeaf),
+            "Invalid proof"
+        );
+        require(msg.value >= args.price, "Insufficient payment");
+        // TODO: Should we only sent the delta between price and sponsor wallet balance? If so, what to do with the rest?
```
bbenligiray, line 128: No, forward the entire `msg.value`

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        // TODO: Will this contract require a withdraw() function?
```
bbenligiray, line 129: Don't see why it would with the above

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+                        args.price
+                    )
+                )
+            )
+        );
+        require(
+            MerkleProof.verify(args.priceProof, args.priceRoot, priceLeaf),
+            "Invalid proof"
+        );
+        require(msg.value >= args.price, "Insufficient payment");
```
bbenligiray, line 127: You're not considering the sponsor wallet balance?

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        // TODO: Should we only sent the delta between price and sponsor wallet balance? If so, what to do with the rest?
+        // TODO: Will this contract require a withdraw() function?
+        Address.sendValue(args.sponsorWallet, msg.value);
```
bbenligiray, line 130: Do this last, this is the least-trusted external call

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+
+        // Deploy the dAPI proxy (if it hasn't been deployed yet)
+        // TODO: check if proxy has been deployed (see: https://github.com/api3dao/airnode-protocol-v1/blob/v2.10.0/contracts/api3-server-v1/proxies/ProxyFactory.sol#L140)
```
bbenligiray, line 133: Yes

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        address proxyAddress = IProxyFactory(proxyFactory).deployDapiProxy(
+            args.dapiName,
+            ""
+        );
+
+        // Update the dAPI with signed API data (if it hasn't been updated recently, what's recently?)
```
bbenligiray, line 139: Recently = last `heartbeatInterval`

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        // TODO: why signed data? shouldn't we instead just call updateBeaconSetWithBeacons()?
```
bbenligiray, line 140: If the Beacons haven't been updated recently, `updateBeaconSetWithBeacons()` doesn't do anything

contracts/Api3Market.sol
```diff
@@ -0,0 +1,151 @@
...
+        bytes32[] memory beaconIds = new bytes32[](args.airnodes.length);
+        for (uint ind = 0; ind < args.airnodes.length; ind++) {
+            beaconIds[ind] = keccak256(
+                abi.encodePacked(args.airnodes[ind], args.templateIds[ind])
+            );
+        }
+        IApi3ServerV1(api3ServerV1).updateBeaconSetWithBeacons(beaconIds);
+
+        // TODO: emit event (include proxyAddress?)
```
bbenligiray, line 149: Yes, and updated sponsor wallet balance

contracts/Api3Market.sol
acenolaza: Right, good catch

contracts/Api3Market.sol
acenolaza: Ok but just to in case the comment wasn't clear my suggestion was more about adding an `if` statement to the `DapiDataRegistry.registerAirnodeSignedApiUrl` function like so:

```
        if (
            keccak256(abi.encodePacked((airnodeToSignedApiUrl[airnode]))) !=
            keccak256(abi.encodePacked((url)))
        ) {
            airnodeToSignedApiUrl[airnode] = url;

            emit RegisteredSignedApiUrl(airnode, url);
        }
```

contracts/Api3Market.sol
acenolaza: My worry was that we might end up with locked funds in the contract but I guess that wouldn't happen if we don't have a `receive() payable` or `fallback() payable` functions (there are some other weird ways to receive funds in this contract like being a recipient of selfdestruct but I don't think is worth considering). Another way to mitigate this could be to forward `address(this).balance` instead of `msg.value` 🤔

contracts/Api3Market.sol
acenolaza: You mean this?

```
require(
            msg.value >= (args.price - args.sponsorWallet.balance),
            "Insufficient payment"
        );
```

or are you referring to deducting balance from previous order when upgrading/downgrading while paid subscription is currently ongoing?

contracts/Api3Market.sol
acenolaza: I just now realized that the "3 months" subscription period from the example is not being stored on-chain 😟 and it feels like this should go into `DapiDataRegistry` along with deviation and heartbeat 🤔

contracts/Api3Market.sol
acenolaza: I guess this means that buyDapi() function needs to also receive signed data for each beacon in order to be able to call [updateBeaconWithSignedData()](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/BeaconUpdatesWithSignedData.sol#L26)

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/ed0f9a544d180d216576740e4a111162790f5bf2

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/62d64f97ea86a5d50641e101c304d67b65a8146c

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/75641b9c080f9b06147aba2473aa0f416310d3c0

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/75641b9c080f9b06147aba2473aa0f416310d3c0

contracts/Api3Market.sol
acenolaza: https://api3workspace.slack.com/archives/C027Y2FAV0S/p1698740982751789?thread_ts=1698659099.119239&cid=C027Y2FAV0S

contracts/Api3Market.sol
```diff
@@ -0,0 +1,314 @@
...
+        Purchase[] storage purchases = dapiToPurchases[dapiNameHash];
+        for (uint256 ind = purchases.length - 1; ind >= 0; ind--) {
+            Purchase storage purchase = purchases[ind];
+            if (
+                block.timestamp >= purchase.start &&
+                block.timestamp < purchase.end
+            ) {
+                index = ind;
+            }
+        }
```
alikonuk1, line 312: This would underflow if `purchases.length` is zero, and also when the loop reaches to zero.
```suggestion
        if (purchases.length == 0) {
            return purchases.length;
        }
        for (uint256 ind = purchases.length; ind > 0; ind--) {
            Purchase storage purchase = purchases[ind - 1];
            if (
                block.timestamp >= purchase.start &&
                block.timestamp < purchase.end
            ) {
                return ind - 1;
            }
        }
        return purchases.length;
```

alikonuk1 (changes requested):

contracts/Api3Market.sol
```diff
@@ -0,0 +1,314 @@
...
+
+        uint256 index = _findCurrentDapiPurchaseIndex(dapiNameHash);
+
+        if (index == 0) {
+            // Scenerio 1: No previous purchases
+            dapiToPurchases[dapiNameHash].push(
+                Purchase(
+                    deviationThresholdInPercentage,
+                    heartbeatInterval,
+                    dapi.price,
```
alikonuk1, line 209: We can use the `priceToPay` here instead of `dapi.price`

alikonuk1 (changes requested):

contracts/Api3Market.sol
```diff
@@ -0,0 +1,350 @@
...
+        dapiFallbackV2 = _dapiFallbackV2;
+        proxyFactory = _proxyFactory;
+        // TODO: should this contract get the Api3ServerV1 address from
+        // DapiDataRegistry or from ProxyFactory contracts instead?
+        api3ServerV1 = _api3ServerV1;
+    }
+
+    // This function must use the 3 Merkle trees to store the data needed for running a managed dAPI
+    function buyDapi(BuyDapiArgs calldata args) external payable override {
+        bytes32 dapiNameHash = keccak256(abi.encodePacked(args.dapi.name));
```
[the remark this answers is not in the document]
alikonuk1: Also it will not work because mapping should be done with `args.dapi.name` ,
Here is how its stored in [DapiFallbakV2.sol](https://github.com/api3dao/dapi-management/blob/214fe632429f982661068dc117bdb66b7659f4b1/contracts/DapiFallbackV2.sol#L205C1-L208C11)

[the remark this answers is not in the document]
acenolaza: That's an issue in `DapiFallbackV2`. Mapping key should be the hash of the `bytes32` encoded string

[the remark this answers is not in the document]
acenolaza: Fixed in commit https://github.com/api3dao/dapi-management/pull/26/commits/c17dae6f65e5f9d783f16d248ccfebdc19e0074e

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/eaa6f41e68c488c4760e84c17271f4296f89701d

contracts/Api3Market.sol
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/c3bec2ebe07e96f5dc94a710ad350806c4a40028

contracts/Api3Market.sol
```diff
@@ -0,0 +1,314 @@
...
+        require(_api3ServerV1 != address(0), "Api3ServerV1 address is zero");
+        hashRegistry = _hashRegistry;
+        dapiDataRegistry = _dapiDataRegistry;
+        proxyFactory = _proxyFactory;
+        // TODO: should this contract get the Api3ServerV1 address from
+        // DapiDataRegistry or from ProxyFactory contracts instead?
+        api3ServerV1 = _api3ServerV1;
+    }
+
+    // This function must use the 3 Merkle trees to store the data needed for running a managed dAPI
```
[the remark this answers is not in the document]
acenolaza: I didn't do it explicitly in Api3Market because the other 2 are being validated by DapiDataRegistry [here](https://github.com/api3dao/dapi-management/blob/170-Api3Market/contracts/DapiDataRegistry.sol#L119-L124) and [here](https://github.com/api3dao/dapi-management/blob/170-Api3Market/contracts/DapiDataRegistry.sol#L219-L224).

alikonuk1 (changes requested):

contracts/Api3Market.sol
```diff
@@ -0,0 +1,371 @@
...
+        }
+        require(msg.value >= updatedPrice, "Insufficient payment");
+    }
+
+    function _findCurrentDapiPurchaseIndex(
+        bytes32 dapiNameHash
+    ) private view returns (bool found, uint256 index) {
+        Purchase[] storage purchases = dapiToPurchases[dapiNameHash];
+        if (purchases.length > 0) {
+            for (uint256 ind = purchases.length - 1; ind >= 0; ind--) {
```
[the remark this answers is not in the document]
acenolaza: While that makes sense, I cannot get it to underflow. Maybe because of the version of solidity we are using which prevents from underflowing. This could be fixed by just using `int256` instead of `uint256` for the `ind` variable or WDYT we should do?

Here's an example that I've used on remix:

```
// SPDX-License-Identifier: GPL-3.0
pragma solidity 0.8.20;

contract Test {
    mapping(bytes32 => Purchase[]) public dapiToPurchases;

    struct Purchase {
        uint256 deviationThreshold;
        uint256 heartbeatInterval;
        uint256 price;
        uint256 duration;
        uint256 start;
        uint256 end;
    }

    constructor() {
        dapiToPurchases[0x9952ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db7].push(Purchase(1 ether, 1 days, 10 ether, 30 days, block.timestamp, block.timestamp + 30 days));
        dapiToPurchases[0x9952ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db7].push(Purchase(2 ether, 1 days, 5 ether, 30 days, block.timestamp + 30 days, block.timestamp + 60 days));
    }

    function _findCurrentDapiPurchaseIndex(
        bytes32 dapiNameHash
    ) external view returns (bool found, uint256 index) {
        Purchase[] storage purchases = dapiToPurchases[dapiNameHash];
        if (purchases.length > 0) {
            for (uint256 ind = purchases.length - 1; ind >= 0; ind--) {
                Purchase storage purchase = purchases[ind];
                if (
                    block.timestamp >= purchase.start &&
                    block.timestamp < purchase.end
                ) {
                    found = true;
                    index = ind;
                    break;
                }
            }
        }
    }
}
```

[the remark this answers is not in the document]
alikonuk1: If you give different bytes32 values like this 
```    
constructor() {
        dapiToPurchases[0x8882ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db8].push(Purchase(1 ether, 1 days, 10 ether, 30 days, block.timestamp, block.timestamp + 30 days));
        dapiToPurchases[0x9952ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db7].push(Purchase(2 ether, 1 days, 5 ether, 30 days, block.timestamp + 30 days, block.timestamp + 60 days));
    }
```
it will not revert on the index 0 (if you call it with `0x8882ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db8`). But if you start on index 1, it will start to revert (if you call it with `0x9952ba9fa4a96966154d47c3df79b8296db7225bc6685e414170b2c7cd4d1db7`).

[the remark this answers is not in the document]
acenolaza: Right, I'll just revert to previous for loop then. Fixed in https://github.com/api3dao/dapi-management/pull/26/commits/1b825657589ad65535e40a2bf0be2e2eb2892f2c

contracts/Api3Market.sol
```diff
@@ -157,6 +150,8 @@ contract Api3Market is IApi3Market {
         );
 
         // This is done last because it is less trusted than other external calls
+        // TODO: in the case of downgrade/upgrade we must fund sponsorWallet with
```
alikonuk1, line 154: Yes I think `_processPayment()` should return an updated amount and we should use that amount to send to the sponsor wallet. And also deduct it from the `msg.value` and return the amount to the caller.

contracts/Api3Market.sol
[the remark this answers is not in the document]
alikonuk1: It seems like the fix got reverted at some point.

contracts/Api3Market.sol
```diff
@@ -0,0 +1,532 @@
...
+                // Scenario 3: New purchase is upgrade
+                uint256 currentOverlapDuration = current.end - block.timestamp;
+                updatedPrice -=
+                    (currentOverlapDuration * current.price) /
+                    current.duration;
+                dapiToPurchases[dapiNameHash].push(
+                    Purchase(
+                        updateParams.deviationThresholdInPercentage,
+                        updateParams.heartbeatInterval,
+                        dapi.price,
```
alikonuk1, line 261: Why don't we use `updatedPrice` here instead of `dapi.price`

contracts/Api3Market.sol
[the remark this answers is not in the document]
acenolaza: I realized that dapi-management frontend would require to get the list of fallbacked dAPI names and not the hashes, that's why I ended up reverting the change on DapiFallbackV2 https://github.com/api3dao/dapi-management/pull/26/commits/f438ee3188295a7f1143c4170587f68cdab5137b

contracts/Api3Market.sol
acenolaza: https://github.com/api3dao/dapi-management/pull/26/commits/efb2a75c1e1d6eada336daa6a8c93db277735fb3

contracts/Api3Market.sol
acenolaza: Because I think we should store the actual cost of the dAPI instead of what the user payed (price minus remainder from previous subscription). This is because we might need to calculate the reminder again in the future in case we upgrade the upgrade

contracts/Api3Market.sol
```diff
@@ -0,0 +1,563 @@
...
+            }
+        }
+    }
+
+    /// @notice Checks if the dAPI has been fallbacked
+    /// @dev Checks if the dAPI is in the list of fallbacked dAPIs in the
+    /// DapiFallbackV2 contract
+    /// @param dapiNameHash Hash of the dAPI name.
+    /// @return isFallbacked True if dAPI name is in the list
+    function _isFallbacked(
```
acenolaza, line 379: Should we move this into a modifier named something like `onlyNotFallbacked` on `DapiFallbackV2` instead of having this function here?

contracts/Api3Market.sol
```diff
@@ -0,0 +1,563 @@
...
+            args.signedApiUrlProofs
+        );
+
+        // Store the actual data used to derive each beaconId (airnode address
+        // and templateId). If more than one is provided in the arguments list
+        // then each beaconId will be used to derive the beaconSetId
+        bytes32 dataFeedId = _registerDataFeed(args.beacons);
+
+        // Add the dAPI to the DapiDataRegistry for managed data feed updates
+        // If purchase if future downgrade, then a worker needs to call the
```
alikonuk1, line 136: How will the worker listen the pending purchases? Will it call `readCurrentAndPendingPurchases()` function periodically or should we emit some event?

alikonuk1 (changes requested):

contracts/Api3Market.sol
alikonuk1: But wouldnt we need to check it from the Api3Market again?

contracts/Api3Market.sol
```diff
@@ -0,0 +1,563 @@
...
+            Purchase storage current = dapiToPurchases[dapiNameHash][index];
+            Purchase storage pending = current;
+            uint256 purchasesLength = dapiToPurchases[dapiNameHash].length;
+            if (purchasesLength > 1 && index == purchasesLength - 2) {
+                pending = dapiToPurchases[dapiNameHash][purchasesLength - 1];
+            }
+
+            require(
+                current.purchasedAt < block.timestamp - 1 days &&
+                    current.purchasedAt < block.timestamp - 1 days,
```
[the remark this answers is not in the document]
acenolaza: Good catch 👏🏻

contracts/Api3Market.sol
acenolaza: Right, I was still thinking about the merge of the 2 contracts but if we leave them separate then at most we can just move the function to DapiFallbackV2 and call it from Api3Market

contracts/Api3Market.sol
acenolaza: I guess both are valid options, it could listen (or fetch) to `BoughtDapi` events to keep track of downgrades by calling `readCurrentAndPendingPurchases()`. Are you thinking that it would be necessary to emit a specific event for downgrades only?

alikonuk1: I don't have strong opinions about it, what you described sounds good.

alikonuk1 (approved):

PR #27, opened 2023-10-30
Endpoint for signing a merkle root

### LINKED ISSUE #16: Add an endpoint for signing a merkle root

pieterdub:
Closes #16

Add nextjs endpoint to sign a merkle root.

Using `multi-manager sig` as reference sign the merkle tree using the static values that is been provided.

endpoint: `localhost:3000/api/sign-merkle-root`
body: 
```
{
        "signature": "0xfe2cd37f6922ecad2e6277165caf65b0a1ed8b854e21ceac45c8f34ed60e3a1150ed6957d7f3f00d40b9f63ad3016051d18513dfeb95b81f8ff362a7017440a11c",
        "address": "0x80efDd3bB15F2108C407049C5575490858800D47",
        "tree": "DAPI_FALLBACK"
}
```


frontend/pages/api/sign-merkle-root/index.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface CurrentHash {
+  timestamp: number;
+  hash: string;
+  signatures: string[];
+  merkleTreeValues: {
+    values: string[][];
+  };
+}
+
+export interface SignMerkleRootRequest {
```
mcoetzee, line 24: Let's use Zod here to validate the request body:
```ts
const requestBodySchema = z.object({ ... });
```
And then we can safeParse:
```ts
const parseResult = requestBodySchema.safeParse(req.body);
if (!parseResult.success) {
  return res.status(400).send(...
}
```

frontend/pages/api/sign-merkle-root/index.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const merkleTreeConfig = {
+  [MerkleTrees.DAPI_FALLBACK]: {
+    fallbackTreeValidator: validateTreeRootSignatures,
+    subfolder: 'dapi-fallback-merkle-tree-root',
+  },
+};
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
```
mcoetzee, line 46: I think we should ensure we are receiving a POST:
```ts
  if (req.method !== 'POST') {
    return res.status(405).send('Method not allowed');
  }
```

mcoetzee (changes requested):

frontend/pages/api/sign-merkle-root/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  if (hashSigners.hashSigners.indexOf(address) === -1) {
+    return res.status(404).send(`Address not part of hash signers for tree ${tree}`);
+  }
+
+  const merkleTree = StandardMerkleTree.of(currentHash.merkleTreeValues.values, ['bytes32', 'bytes32', 'address']);
+  const root = ethers.utils.arrayify(merkleTree.root);
+
+  // Set a new signature on the index belonging to the signing address
+  currentHash.signatures[hashSigners.hashSigners.indexOf(address)] = signature;
```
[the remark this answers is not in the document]
pieterdub: This is being validated on the next step. If this is removed it will validate the previous signature for this address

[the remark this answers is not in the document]
mcoetzee: Ah I'm with you now 👍

frontend/pages/api/sign-merkle-root.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method !== 'POST') {
+    return res.status(405).send('Method not allowed');
+  }
+
+  const parseResult = requestBodySchema.safeParse(req.body);
+  if (!parseResult.success) {
+    return res.status(400).send(`Invalid body:\n, ${JSON.stringify(parseResult.error.format())}`);
```
mcoetzee, line 55: Minor: Hmm would probably make more sense (in the context of a API response) to send back JSON here rather than a formatted string:
```ts
return res.status(400).json(parseResult.error);
```

frontend/pages/api/sign-merkle-root.ts
```diff
@@ -0,0 +1,108 @@
...
+  exec('yarn prettier');
+
+  return res.status(200).send('Successfully signed root');
+}
+
+interface JSONObject {
+  [x: string]: JSONValue;
+}
+
+type JSONValue = string | number | boolean | JSONObject | string[] | string[][];
```
mcoetzee, line 104: I assume all this is copy-pasta'd from somewhere, I'm just struggling to see the need for it if all we need is:
```ts
writeFileSync(currentHashPath, JSON.stringify(validatedCurrentHash));
```

mcoetzee (changes requested):
Looking good 👍

mcoetzee (approved):
Using `exec('yarn prettier');` simplified things nicely 🥂

PR #31, opened 2023-10-31
Add initial dAPI fallback merkle tree page

### LINKED ISSUE #21: Add initial dAPI fallback merkle tree page

mcoetzee:
Closes #21 

### What does this change?
- adds the dAPI fallback merkle tree page (the Sign Root and View Tree Diff buttons do nothing at this stage)
- adds some lower level components that can used to when building the other MT pages
- adds a `Table` component via `shadcn-ui` with the added ability to make the table header sticky
- tweaks the root layout to better handle horizontal scrolling
- turns on experimental scroll restoration when clicking back button (via nextjs config)

### Screenshots
fallback-tree-2


frontend/next.config.js
```diff
@@ -1,6 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: false,
+  experimental: {
+    scrollRestoration: true,
```
pieterdub, line 5: Nifty 👏

frontend/pages/merkle-trees/dapi-fallback.tsx
```diff
@@ -1,5 +1,120 @@
...
+    values: z.array(z.tuple([z.string(), z.string(), z.string()])),
+  }),
+});
+
+const signersSchema = z.object({
+  hashSigners: z.array(z.string()),
+});
+
+export function getServerSideProps() {
+  const currentTreePath = join(process.cwd(), '../data/dapi-fallback-merkle-tree-root/current-hash.json');
```
pieterdub, line 26: Would be great if we can abstract the fetching and parsing of the files to a util which the api endpoint could also use.

```
function getMerkleTreeMetadata(treeFolder: TreeSubfolder){
    const currentTreePath = join(process.cwd(), '../data/dapi-fallback-merkle-tree-root/current-hash.json');
    const signersPath = join(process.cwd(), `../data/${treeSubfolder}/hash-signers.json`);
    const currentTree = JSON.parse(readFileSync(currentTreePath, 'utf8'));
    const signers = JSON.parse(readFileSync(signersPath, 'utf8'));

    return {currentTree, singers};|
}
```

Thoughts?

pieterdub (commented):
LGTM, just a question

frontend/pages/merkle-trees/dapi-fallback.tsx
mcoetzee: Yeah maybe a function per file, because the `getServerSideProps()` function of each page will also need to read the `previous-hash.json` file.

frontend/lib/server/file-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+});
+
+const signersSchema = z.object({
+  hashSigners: z.array(z.string()),
+});
+
+type MerkleTreeData = z.infer<typeof merkleTreeSchema>;
+type TreeFile = 'current-hash.json' | 'previous-hash.json';
+
+export function readTreeDataFrom(options: { subfolder: TreeSubFolder; file: TreeFile }): {
```
pieterdub, line 25: Not sure I follow the use of this function

frontend/lib/server/file-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  path: string;
+  data: MerkleTreeData;
+};
+
+export function readTreeDataFrom<T>(options: { subfolder: TreeSubFolder; file: TreeFile; schema: z.ZodSchema<T> }): {
```
pieterdub, line 30: Also this

frontend/lib/server/file-utils.ts
mcoetzee: This is called function overloading, and these first two are signature overloads: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

pieterdub (approved):
🚀

PR #32, opened 2023-10-31
Function to be able to switch back from fallback on a specific dAPI

acenolaza:
While thinking about merging (inherit) `DapiFallbackV2` with `Api3Market` I remembered that `DapiFallbackV2` must be kept funded in order to be able to send funds to the self funded sponsor wallet in the case of calling `executeDapiFallback()`. It seems to me that it might not be a good idea to keep and transfer funds from the same contract that receives payments for managed dAPIs. Therefore keeping the separation of concerns seems to still be a good idea after all and I even applied the same logic for the new list of dAPIs for which `executeDapiFallback()` has been called. Having this list in `DapiDataRegistry` did not look appropriate and everything fallback related should be in the `DapiFallbackV2`. 

With that said, I see two main drawbacks:
1. `DapiFallbackV2` will require the `addDapi` and `removeDapi` roles to be granted
2. Market UI will need to use a separate RPC call to fetch the list of fallbacked dAPIs

I've created this branch out of main branch because I ended up not touching the Api3Market contract. Please share your thoughts  🙏🏻  (as usual I left TODOs with questions/concerns)


contracts/DapiDataRegistry.sol
```diff
@@ -99,9 +92,13 @@ contract DapiDataRegistry is
...
         require(_api3ServerV1 != address(0), "Api3ServerV1 address is zero");
-        registrarRole = _deriveRole(
+        dapiAdderRole = _deriveRole(
+            _deriveAdminRole(manager),
+            DAPI_ADDER_ROLE_DESCRIPTION
+        );
+        dapiRemoverRole = _deriveRole(
             _deriveAdminRole(manager),
-            REGISTRAR_ROLE_DESCRIPTION
+            DAPI_REMOVER_ROLE_DESCRIPTION
```
acenolaza, line 101: I felt lazy for not using a separate role and now I had an excuse for doing so 😄

contracts/DapiFallbackV2.sol
```diff
@@ -143,9 +165,67 @@ contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
...
+    // This function requires the root and a proof from the dAPI management
+    // Merkle tree and that the fallback dAPI function was previously executed.
+    // TODO: Only the contract owner can execute this function to switch back to
+    // previous payed dAPI subscription. Why anyone can call executeDapiFallback?
+    // We used to have a onlyByDapiFallbackExecutorWithInd modifier to have only
+    // a group of people be able to execute this function. I know that data from
+    // merkle trees has been reviewed and signed by other people but anyone that
+    // gets a hold of the json files with the merkle trees or by just clicking on
+    // the dapi management UI can still switch to fallback at any given time?
+    function revertDapiFallback(
```
acenolaza, line 188: natspec comments are missing but will be added once and if we decide this is a valid approach

contracts/DapiFallbackV2.sol
```diff
@@ -143,9 +165,67 @@ contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
...
+    function getFallbackedDapis()
+        external
+        view
+        override
+        returns (bytes32[] memory dapis)
+    {
+        dapis = fallbackedDapis.values();
+
+        // TODO: This function is intended to be used by the UI to still be able to display fallbacked dAPIs
+        //       Should we store the dataFeedId and sponsorWallet used on executeDapiFallback? update params
```
alikonuk1, line 225: Would it be better if UI fetch those values from Nodary for fallbacked dapis?

alikonuk1 (changes requested):

contracts/DapiFallbackV2.sol
```diff
@@ -143,9 +165,67 @@ contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
...
+
+        IDapiDataRegistry(dapiDataRegistry).removeDapi(args.dapiName);
+
         emit ExecutedDapiFallback(args.dapiName, args.dataFeedId, msg.sender);
     }
 
+    // This function requires the root and a proof from the dAPI management
+    // Merkle tree and that the fallback dAPI function was previously executed.
+    // TODO: Only the contract owner can execute this function to switch back to
+    // previous payed dAPI subscription. Why anyone can call executeDapiFallback?
```
alikonuk1, line 182: I think the thought process was to keep fallback flow to be less dependent on just owners but giving more flexibility to anyone who has the right values to execute. But I see your point, some malicious actor can constantly execute fallbacks.

contracts/DapiFallbackV2.sol
```diff
@@ -143,9 +165,67 @@ contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
                 msg.sender
             );
         }
+
+        require(
+            fallbackedDapis.add(args.dapiName),
+            "dAPI fallback already executed"
+        );
+
+        IDapiDataRegistry(dapiDataRegistry).removeDapi(args.dapiName);
```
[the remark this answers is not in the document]
acenolaza: Fixed in https://github.com/api3dao/dapi-management/pull/32/commits/2f253ab2ed5633472efaab527d7f79768baa8fa5

contracts/DapiFallbackV2.sol
```diff
@@ -143,9 +165,67 @@ contract DapiFallbackV2 is Ownable, IDapiFallbackV2 {
...
+        uint256 deviationThresholdInPercentage,
+        int224 deviationReference,
+        uint32 heartbeatInterval,
+        bytes32 root,
+        bytes32[] calldata proof
+    ) external override onlyOwner {
+        require(
+            fallbackedDapis.remove(dapiName),
+            "dAPI fallback has not been executed"
+        );
```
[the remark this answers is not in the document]
acenolaza: I did add the `require` check on [`remove()`](https://github.com/api3dao/dapi-management/blob/DapiFallbackV2-DapiDataRegistry/contracts/DapiDataRegistry.sol#L268) but I'm not so sure we need it in [`add()`](https://github.com/api3dao/dapi-management/blob/DapiFallbackV2-DapiDataRegistry/contracts/DapiDataRegistry.sol#L235) because this is the intended way of updating the update parameters for an already added dAPI ([upgrade/downgrade](https://github.com/api3dao/dapi-management/blob/DapiFallbackV2-DapiDataRegistry/contracts/DapiDataRegistry.sol#L182-L185))

contracts/DapiFallbackV2.sol
acenolaza: Right, I was thinking the same thing 👍🏻 
Fixed in https://github.com/api3dao/dapi-management/pull/32/commits/89ba7d0c3bb0367422abb9912160c613e1875355

contracts/DapiFallbackV2.sol
acenolaza: Right, and also since DapiDataRegistry.addDapi() and DapiDataRegistry.removeDapi() are behind access control then it seems wrong to me that DapiFallbackV2.executeDapiFallback() and DapiFallbackV2.revertDapiFallback() are not behind some access control check too.

acenolaza:
@alikonuk1 Since there seems to be no major objections to this approach over the merging of this contract with Api3Market contract then I'm setting this as "Ready for review"

contracts/DapiFallbackV2.sol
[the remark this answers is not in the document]
alikonuk1: My bad, its more clear now.

contracts/DapiFallbackV2.sol
alikonuk1: Should we do this on this PR or add it later?

acenolaza: TBH I'm not sure. Maybe it's OK that anyone can switch to fallback but I'm really uncomfortable letting anyone switch a dAPI to any data feed in the merkle tree 🤷🏻 WDYT?

alikonuk1: How does this sound, lets put some access control and give access to larger group of people than signers. This way, we won't need just signers to execute but will also have other "less-trusted" executors who can step up and execute a fallback in times of emergency.

acenolaza: Right, similar to the `dapiFallbackExecutors` list we used to have. I guess it only makes sense that dAPI management merkle tree root signers are allow to execute / revert dAPIs fallback. I'm thinking thinking that fallback merkle tree root signers should be left out from the executors list (this used to be the way in V1)

alikonuk1: Added necessary checks in [d472736](https://github.com/api3dao/dapi-management/pull/32/commits/d472736082ef684e4a6a96913bf83c87d445ddf4)

alikonuk1 (approved):
LGTM

PR #39, opened 2023-11-09
Add merkle tree diff viewer (dAPI fallback page)

### LINKED ISSUE #24: Add merkle tree diff viewer (dAPI fallback page)

mcoetzee:
Closes #24 

### What does this change?
- generates a `git diff` when the page loads and uses the https://diff2html.xyz/ library to parse and display the diff
- adds a `<Tabs />` component via `shadcn-ui` and uses it to display the _Tree Values_ and new _Tree Diff_ 
- adds a `<Toggle />` component via `shadcn-ui` and uses it to toggle between viewing the tree diff in "split" and "unified" modes. It defaults to "split" mode and the user preference gets saved in local storage
- extracts the tree diff viewer as a reusable component (for the other MT pages)
- updates the `dapi-fallback-merkle-tree-root/current-hash.json` to store the _dAPI names_ as plain text so that the diff viewer is easier to read
- formats the plain text _dAPI names_ before constructing the `StandardMerkleTree`s in all the places (scripts, sign endpoint, dAPI fallback page)
- further updates the sign MT script:
  - updates the MT names received as argument to align with the names used to create the hash types, i.e.  `'dAPI fallback Merkle tree', 'dAPI management Merkle tree', 'dAPI pricing Merkle tree', 'Signed API URL Merkle tree'`
  - authorises the signer by checking if they are included in the root signers list (`hash-signers.json`)
  - removes the unused `chainId` argument
  - throws an error if the `current-hash.json` file doesn't exist
- updates to `scripts` readme to reflect above changes

### Screenshots
Split mode
split-view
Unified mode
unified-view


mcoetzee:
@alikonuk1 FYI I have made some updates to the `signMerkleTree.js` script (as outlined in the PR description). Feel free to leave comments on those changes 👍

frontend/pages/api/sign-merkle-root.ts
```diff
@@ -60,7 +65,7 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
   // for every signer check if the signature is valid and if not replace it with "0x"
   const validatedRootSignatures = validateTreeRootSignatures(root, currentHash.signatures, hashSigners);
 
-  const validatedCurrentHash = { ...currentHash, signatures: validatedRootSignatures };
+  const validatedCurrentHash = { ...currentHash, hash: merkleTree.root, signatures: validatedRootSignatures };
```
pieterdub, line 68: The hash should not be updated here. As I understand it the hash is the hash of `hashType`, treeRoot and `timestamp` or am I mistaken?
 I noticed an issue with this while doing my current task. The schema of the `readTreeDataFrom` method should include the `hash` and `timestamp`.

mcoetzee: > The hash should not be updated here.

The `signMerkleTree.js` script assigns the `hash`: https://github.com/api3dao/dapi-management/blob/5caf26162bf6d4be1c5bcb2a25f736c8b7f5eda2/scripts/signMerkleTree.js#L67
So it makes sense for the endpoint to be in sync with the script (the readme describes a workflow where the user updates the values and signs afterwards: https://github.com/api3dao/dapi-management/blob/5caf26162bf6d4be1c5bcb2a25f736c8b7f5eda2/scripts/README.md).


> As I understand it the hash is the hash of hashType, treeRoot and timestamp or am I mistaken?

The `signMerkleTree.js` script assigns the tree root as the `hash` value which is why I am under the impression that the value should be the tree root. @alikonuk1 is @pieterdub correct and the script perhaps out of date?


> I noticed an issue with this while doing my current task. The schema of the readTreeDataFrom method should include the hash and timestamp.

You add things to schemas as you start using them, so it's not an issue if something is missing that hasn't been used before.

alikonuk1: This script looks okay to me.

> As I understand it the hash is the hash of hashType, treeRoot and timestamp or am I mistaken?

The hash that is mentioned in the json file is the tree root actually. My mistake, naming of them are bit confusing, now I can see it.

pieterdub (approved):
LGTM 🚀

frontend/pages/api/sign-merkle-root.ts
```diff
@@ -69,3 +74,16 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
 
   return res.status(200).send('Successfully signed root');
 }
+
+function createMerkleTree(type: TreeType, values: string[][]) {
```
pieterdub, line 78: This feels like it should be merged with the above switch statement. Perhaps:
```
interface TreeConfig {
  subfolder: TreeSubFolder;
  merkleTree: Tree
}

function getTreeConfig(type: TreeType): TreeConfig {
  switch (type) {
    case 'dapi-fallback':
      return {subfolder: 'dapi-fallback-merkle-tree-root', merkleTree: createDapiFallbackMerkleTree(values)};
    case 'dapi-management':
      return {subfolder: 'dapi-management-merkle-tree-root', merkleTree: createDapiManagementMerkleTree(values)};
    case 'dapi-pricing':
      return {subfolder: 'dapi-pricing-merkle-tree-root', merkleTree: createDapiPricingMerkleTree(values)};
    case 'signed-api-url':
      return {subfolder: 'signed-api-url-merkle-tree-root', merkleTree: createSignedApiUrlMerkleTree(values)};
  }
}
```

Thoughts? I need to map the string provided to the hashType string as well

mcoetzee: > I need to map the string provided to the hashType string as well

Iirc the `TreeType` should be aligned with the hashType string, so you would be able to just:
```ts
const hashType = `${treeType} root`;
```

mcoetzee: That being said, I can merge the two 👍

frontend/pages/api/sign-merkle-root.ts
bbenligiray: I think I named that `hash` field. The reasoning is that the HashRegistry contract allows any kind of hash to be registered, not necessarily only a Merkle tree root. So the JSON schema is partially unopinionated about this (for the first `timestamp`, `hash`, `signatures` fields), which would potentially allow us to extend this repo to support other kinds of hashes in the future.

PR #60, opened 2023-11-22
Adds basic deployment script for local frontend development

acenolaza:
Closes #195

### What does this PR changes?
1. Adds a basic deployment script that allows to deploy all contracts to a local running node
2. All contracts use the deterministic deployment feature from hardhat-deploy plugin
3. `HashRegistry` constructor was modified to be able to deploy it using deterministic deployment

I was thinking that it might be useful to set up the roles as well but that might be trying to do too much and may be better to as a separate step/script


deploy/1_deploy.js
```diff
@@ -0,0 +1,114 @@
...
+
+    dapiFallbackManagers = [dapiFallbackManager1, dapiFallbackManager2, dapiFallbackManager3];
+  } else {
+    ownableCallForwarderAddress = airnodeProtocolV1.references.OwnableCallForwarder[network.config.chainId.toString()];
+    accessControlRegistryAddress =
+      airnodeProtocolV1.references.AccessControlRegistry[network.config.chainId.toString()];
+    api3ServerV1Address = airnodeProtocolV1.references.Api3ServerV1[network.config.chainId.toString()];
+    proxyFactoryAddress = airnodeProtocolV1.references.ProxyFactory[network.config.chainId.toString()];
+
+    // TODO: Is this the right list of manager accounts?
```
alikonuk1, line 56: We might want to extend this access to a broader list of individuals. I see no problem with more people executing it, as it's relatively secure (requiring signatures from the right people) and issues are easy to spot. But starting with the dAPI Management hash signers should be okay initially.

alikonuk1 (approved):

deploy/1_deploy.js
```diff
@@ -0,0 +1,114 @@
+const airnodeProtocolV1 = require('@api3/airnode-protocol-v1');
+const { hashSigners: dapiHashSigners } = require('../data/dapi-management-merkle-tree-root/hash-signers.json');
```
mcoetzee, line 2: Did you mean to import the signers from `../data/dapi-fallback-merkle-tree-root/hash-signers.json`?

acenolaza: I know that this looks kind of weird 😅 but we basically want to allow people other than the ones that said "these are the valid fallback data feed ids" (dapi fallback hash signers). So for example if Bedirhan, Mertcan and Burak are the ones signing the `dapi-fallback-merkle-tree-root` hash then we want other people to be ones that can actually execute fallbacks with that data (let say, you, Ali and me)

mcoetzee: Ah ok thanks for the explanation 👍

PR #73, opened 2023-11-27
HashRegistry metadata verification scripts

acenolaza:
Closes [#197](https://github.com/api3dao/tasks/issues/197)

### What does this change?

1. Adds a new hardhat-deploy script to keep track of deployed contract addresses and deployment block numbers (2_document.js)
2. Adds a script that iterates over the list of chains where contracts have been deploy. For each, it starts a new process where metadata values are verified on chain (HashRegistry contract). This is mainly because I wanted to levarage the augmented hardhat runtime environment that hardhat-deploy plugin offers in the scripts. The thing is that HRE requires a network to be set.
3. Refactors test.yml Github workflow to be consistent with other workflows used in other repos


alikonuk1 (approved):
LGTM👍

scripts/verify-hash-registry.js
```diff
@@ -0,0 +1,100 @@
...
+
+async function verify(hashRegistry, hashSigners, hashType, hash, timestamp, signatures) {
+  const onChainHashSigners = await hashRegistry.getSigners(hashType);
+  if (!hashSigners.every((signer, index) => signer === onChainHashSigners[index])) {
+    throw new Error('On-chain signers mismatch');
+  }
+  const message = hre.ethers.utils.arrayify(
+    hre.ethers.utils.solidityKeccak256(['bytes32', 'bytes32', 'uint256'], [hashType, hash, timestamp])
+  );
+  if (
```
mcoetzee, line 25: Not really an issue, but we've got a small overlap here with [frontend/tests/merkle-tree-files.test.ts#L28](https://github.com/api3dao/dapi-management/blob/main/frontend/tests/merkle-tree-files.test.ts#L28):
```ts
  // TODO: Enable when the time comes to sign
  describe.skip('signatures', () => {
    it('has keys that match the hash signers array', () => {
      expect(Object.keys(dapiFallbackTree.signatures)).toEqual(dapiFallbackTreeSigners.hashSigners);
    });

    it('only contains verified signatures', () => {
      const hashToSign = deriveTreeHash(
        'dAPI fallback Merkle tree root',
        dapiFallbackTree.hash,
        dapiFallbackTree.timestamp
      );

      forEach(dapiFallbackTree.signatures, (signature, address) => {
        expect(ethers.utils.verifyMessage(hashToSign, signature)).toEqual(address);
      });
    });
  });
```

Do you think it's necessary to verify the signatures here again?

scripts/verify-metadata.js
```diff
@@ -0,0 +1,28 @@
...
+  for (const [, network] of Object.entries(references.chainNames)) {
+    const childProcess = spawn('node', ['./scripts/verify-hash-registry.js'], {
+      stdio: 'inherit',
+      env: {
+        ...process.env,
+        HARDHAT_NETWORK: network,
+      },
+    });
+
+    childProcess.on('close', (code) => {
```
mcoetzee, line 14: Hmm wouldn't this `main` function complete without waiting for these callbacks? `spawn` is async, so I think the `main` function will always exit as soon as it has iterated over the chains (in other words, this script will never produce an exit code of 1). Or am I missing something?

scripts/verify-hash-registry.js
acenolaza: Interesting, I see that this tests are a bit different tho:
1. In the first `it` block you are comparing signatures in the current-hash.json against signers in the hash-signers.json. In the verify script I compare the signatures against the on-chain signers list.
2. Similarly, in the second `it` block you are basically comparing that the key (signer address) was the account that actually produced the value (signature). In the script I check that the on-chain signers were the ones that produced those signatures.
Bottom line is that I think these tests can be complementary with the local json files vs on-chain values verification. WDYT?

scripts/verify-metadata.js
acenolaza: I been able to get this callback to print "Child process exited with code..." log message before the main finishes while testing against localhost. I guess I might need to test with more networks.

acenolaza: I think [this](https://github.com/api3dao/dapi-management/pull/73/commits/8d4f5dab9214e7e5704eceb86615bdd48ea78367) might be an improvement

scripts/verify-hash-registry.js
mcoetzee: Yeah basically those two tests ^^ together verifies the entries in the `signatures` object against the signers in the `hash-signers.json` file. So if the signers in the `hash-signers.json` matches the on-chain signers, then we know all is good.

But I'm not fussed, it can't hurt to have an additional sanity check in your script.

mcoetzee (approved):

PR #74, opened 2023-11-28
Add dapi-pricing split script and run script in github action

pieterdub:
Closes #63 

When there has been a push made to main we want to run a script on the gh-pages branch. Updating the action on main branch to handle this.

This adds a script to the `gh-pages` branch that will be executed by a github action when there is a push done on main.

The script uses the dapi-pricing `current-hash.json` and loops through the values.

1. Create a object that will be used to write the values to files.
2. This creates a object with all the values on a specific chainId and that object has multiple dapi-names with values that have different deviations applied to the price.

### Metadata
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/metadata.json

### Value 
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/100/USDC-USD.json

https://github.com/api3dao/dapi-management/blob/ed787fb577a2a08f5738586c29cfeae65c617d63/data/dapi-pricing-merkle-tree-root/1/AAPL-USD.json

4. This object has the value and the proof that was generated using `StandardMerkleTree.of` object.
5. Using this object directories are created that will show the correct path needed to fetch the json on github pages. Then it writes the contents of that object to each respective file

```
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/<chainId>/<dapi-name>.json
```

Tested it using a test branch here:
https://github.com/api3dao/dapi-management/tree/ed787fb577a2a08f5738586c29cfeae65c617d63/data/dapi-pricing-merkle-tree-root


mcoetzee (changes requested):
This worked out quite nicely 👌

scripts/split-dapi-pricing.js
```diff
@@ -0,0 +1,58 @@
...
+      const filePath = path.join(dirPath, `${dapiName}.json`);
+      await fs.promises.mkdir(dirPath, { recursive: true });
+      fs.writeFileSync(filePath, JSON.stringify(content, null, 4));
+    }
+  }
+
+  const metadatPath = path.join(__dirname, '..', 'data', 'dapi-pricing-merkle-tree-root', 'metadata.json');
+  fs.writeFileSync(metadatPath, JSON.stringify(metdata, null, 4));
+
+  exec('yarn format');
```
mcoetzee, line 50: Ah I almost forgot to add that I think we will need to await this. You can get an awaitable `execute` fn by:
```ts
import { promisify } from 'util';
import { exec } from 'child_process';

const execute = promisify(exec);
```

scripts/split-dapi-pricing.js
```diff
@@ -0,0 +1,61 @@
...
+    const leaf = { value: item, proof };
+
+    if (!valuesByChainAndDapiName[chainId]) {
+      valuesByChainAndDapiName[chainId] = {};
+    }
+
+    if (!valuesByChainAndDapiName[chainId][name]) {
+      valuesByChainAndDapiName[chainId][name] = [leaf];
+    } else {
+      // Clone the existing array and append the new leaf
```
mcoetzee, line 32: Minor: This comment no longer applies, and can be removed

mcoetzee (approved):
Looking good🥂

PR #75, opened 2023-11-28
Add script to handle dapi-pricing values split

pieterdub:
## THIS PR MERGES INTO GH-PAGES BRANCH

This adds a script to the `gh-pages` branch that will be executed by a github action when there is a push done on main.

The script uses the dapi-pricing `current-hash.json` and loops through the values.

1. Create a object that will be used to write the values to files.
2. This creates a object with all the values on a specific chainId and that object has multiple dapi-names with values that have different deviations applied to the price.

### Metadata
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/metadata.json

### Value 
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/100/USDC-USD.json

https://github.com/api3dao/dapi-management/blob/ed787fb577a2a08f5738586c29cfeae65c617d63/data/dapi-pricing-merkle-tree-root/1/AAPL-USD.json

4. This object has the value and the proof that was generated using `StandardMerkleTree.of` object.
5. Using this object directories are created that will show the correct path needed to fetch the json on github pages. Then it writes the contents of that object to each respective file

```
https://api3dao.github.io/dapi-management/data/dapi-pricing-merkle-tree-root/<chainId>/<dapi-name>.json
```

Tested it using a test branch here:
https://github.com/api3dao/dapi-management/tree/ed787fb577a2a08f5738586c29cfeae65c617d63/data/dapi-pricing-merkle-tree-root


scripts/split-dapi-pricing.js
```diff
@@ -0,0 +1,57 @@
+const fs = require('fs');
```
mcoetzee, line 1: Hmm I'm not convinced that we should be committing code directly to the `gh-pages` branch. It would make things more difficult to maintain.

It would also be nice if these files weren't exposed by GH pages.

What if the code lived on `main` and GH action additionally checked out the files you need to make the script run (from `main`). I'm guessing:
- package.json
- yarn.lock
- scripts folder

And then the workflow could be something like:
- checkout data files
- if changes are detected, checkout the additional files
- run script
- git add data (so we ignore the other files)
- commit and push

pieterdub: Mmm yeah that would make more sense. I will make those changes 👍

PR #83, opened 2023-12-04
Make Merkle tree values more human-readable (on the frontend)

### LINKED ISSUE #71: Make Merkle tree values more human-readable on the frontend
The same rules apply to both the Tree Values and Tree Diff tabs.

**dAPI pricing MT**
- dAPI name: Parse the bytes32
- Chain ID: Display the chain alias instead
- dAPI update parameters: Display the deviation threshold in % and the heartbeat interval in hours (and display the deviation reference if it isn't zero)
- Duration: Display the duration in days
- Price: Already covered by the unit selection at the header

**dAPI management MT**
- dAPI name: Parse the bytes32
- Data feed ID: List the API provider names (If you know the Beacon set ID, you know the Beacon IDs. If you know the Beacon IDs, you know the API providers), complain if any of the Beacon IDs does not match the feed name
- dAPI sponsor wallet address: Just have a ✅ (and complain if it's wrong). You know the xpub of Airseeker-v2 (should be in the dpai-management repo) and you can derive the sponsor wallet address from that and the feed name (https://github.com/api3dao/airseeker-v2/blob/main/src/utils.ts#L30)

**dAPI fallback MT**
- dAPI name: Parse the bytes32
- Data feed ID: List the API provider names (same as above, we expect this to just read "nodary")
- Sponsor wallet address: Just have a ✅ (and complain if it's wrong). You know the xpub of the Nodary Airseeker and can derive the sponsor wallet address from that and the data feed ID (https://github.com/nodaryio/utilities/blob/main/src/index.js#L56)

**Signed API URL MT**
- Airnode address: Display the API provider name, they are all in api-integrations
- Signed API URL: This is already a string, nothing to do here

mcoetzee:
Closes #71 

### What does this change?
- adds the ability to pre-process the `current-hash.json` and `previous-hash.json` files in order to create a `git diff` on the processed files (processed files get saved to the .gitignored `data/.processed/` folder) 
Screenshot 2023-12-05 at 12 24 19
- formats the dAPI management, dAPI pricing, and Signed API URL MT values to make them human-readable
- adds the `dropdown-menu` component via `shadcn-ui` and uses it to add a "View" options menu to all 3 pages that allows you to toggle between viewing the Human-readable/Raw tree values (this takes effect on both the Tree Values and Tree Diff tabs). The menu also allows you to toggle between viewing the tree diff in "side by side" or "unified" mode.
- replaces the addresses with the names of known signers (in the signatures object) when viewing the human-readable tree diff
- adds a toast message to the dAPI management page to inform the user that the CI verifies the Data Feed IDs and the Sponsor Wallet Addresses for the user
- additionally adds tooltips to the dAPI management page's table header to indicate that the CI verifies above data
- extends the Tooltip to accept a `preventCloseOnClick` prop (to make it easier to reuse this behaviour)
- updates the default Tooltip styling (to have a dark bg)
- uses a tooltip rather on the disabled Sign Root button to indicate when a user is not a hash signer
- extracts the logic to get the page props for each MT page
```ts
  const { currentTree, signers, diffResult } = await getMerkleTreeServerSideProps({
    subfolder: 'dapi-management-merkle-tree-root',
    schema: merkleTreeSchema,
    diff: {
      preProcess: !showRawValues,
      preProcessor: (values) => {
        const dapiName = ethers.utils.parseBytes32String(values[0]);
        return [dapiName, getProviders(dapiName), values[2]];
      },
    },
  });
```

### How did you action this task?
W.r.t. toggling between viewing Human-readable and raw tree values, a URL query param was used:
```
http://localhost:3000/merkle-trees/dapi-management?raw=true
```
Setting the query param triggers the `getServerSideProps()` of the page again, where the git diff is then generated accordingly.

### Screenshots
Screenshot 2023-12-04 at 15 36 06

https://github.com/api3dao/dapi-management/assets/747979/a81fa26f-5d88-4ca1-9f10-0eaacc16b3c9

https://github.com/api3dao/dapi-management/assets/747979/aa23fae1-069a-4564-a9df-c3c785c730fe

Screenshot 2023-12-04 at 16 03 05


frontend/components/merkle-tree-elements/index.tsx
```diff
@@ -124,114 +135,19 @@ interface SignerInfoProps {
 }
 
 function SignerInfo(props: SignerInfoProps) {
-  const triggerRef = useRef<HTMLButtonElement>(null);
   return (
-    <Tooltip delayDuration={0}>
+    <Tooltip delayDuration={0} preventCloseOnClick>
```
pieterdub, line 139: Nice 👍

frontend/pages/merkle-trees/dapi-pricing.tsx
```diff
@@ -83,49 +94,117 @@ export default function DapiPricingTree(props: Props) {
...
+
+  return devReference ? `${deviation}, ${devReference}, ${heartbeat}` : `${deviation}, ${heartbeat}`;
+}
+
+function formatDuration(seconds: string) {
+  return round(parseInt(seconds, 10) / 24 / 60 / 60, 2).toString();
+}
+
+let aliases: Record<string, string>;
+function getChainAlias(chainId: string) {
```
pieterdub, line 201: Seems like a better place for this would be `utils`

pieterdub (commented):
Nice, good work. Just a single comment. There might be a few conflicts with my PR as I see you have remove the use of the merkle tree constructor in the pricing page. I am also not sure if we need to add any more changes as that page won't display the data table

frontend/pages/merkle-trees/dapi-pricing.tsx
mcoetzee: If we need to use this function somewhere else, then we can move it to somewhere more global. As a general rule, I keep things local when not needed elsewhere 👍

mcoetzee:
> There might be a few conflicts with my PR as I see you have remove the use of the merkle tree constructor in the pricing page.

It will be straight forward to resolve.

> I am also not sure if we need to add any more changes as that page won't display the data table.

We're merging these dAPI Pricing page changes, so that we have them in our git history in case we need to get the code back in the future 👍

mcoetzee:
Following this [discussion](https://api3workspace.slack.com/archives/C061WEUEXRP/p1701850629767969):
- the Sponsor Wallet Address column has been removed from the human-readable views (dAPI Management)
- the toast message and tooltips wrt the CI have been updated to make it more obvious that the user needs to check the CI to make sure the verification was a success.
Screenshot 2023-12-06 at 12 50 50


Screenshot 2023-12-06 at 12 51 07

PR #79, opened 2023-11-29
Dapi man/57

metobom:

PR #91, opened 2023-12-06
Extend the CI to verify the Data Feed IDs, Sponsor Wallet Addresses, and more

### LINKED ISSUE #50: Extend the CI to verify the Data Feed IDs and Sponsor Wallet Addresses (dAPI management)

mcoetzee:
Closes #50 

### What does this change?
- deletes the `frontend/tests/merkle-tree-files.test.ts` and adds the equivalent assertions to:
```
scripts/verification/
  |- verify-dapi-management-data.js
  |- verify-dapi-pricing-data.js
  |- verify-signed-api-url-data.js
```
The `deriveSponsorWalletAddress` function for some reason runs very slowly when run via Jest (over 1s per call vs 10-20ms when run directly via Node). 
- renames the `scripts/verify-metadata.js` -> `scripts/verify-data.js` and extends it to also call the verify functions noted above
Screenshot 2023-12-07 at 10 56 17

- extends `verify-dapi-management-data.js` to verify Data Feed IDs, Sponsor Wallet Addresses, and the `previous-hash.json` hash value (the last being a sanity check for the tree diff)
- extends `verify-signed-api-url-data.js` to verify each Airnode Address has a mapping to API providers, and the `previous-hash.json` hash value (the last being a sanity check for the tree diff)
- extracts the logic to derive data feed IDs and sponsor wallet addresses from `generate-dapi-management-mt.js` to `scripts/utils.js` and uses these function in both `generate-dapi-management-mt.js` and `verify-dapi-management-data.js`


mcoetzee:
FYI @metobom 
> - extracts the logic to derive data feed IDs and sponsor wallet addresses from `generate-dapi-management-mt.js` to `scripts/utils.js` and uses these function in both `generate-dapi-management-mt.js` and `verify-dapi-management-data.js`

Feel free to leave comments on these changes.

metobom:
> Feel free to leave comments on these changes.

Makes sense since they will be used in multiple places.

scripts/verification/verify-dapi-management-data.js
```diff
@@ -0,0 +1,75 @@
...
+  getDapiManagementHashType,
+  deriveDataFeedId,
+  deriveSponsorWalletAddress,
+} = require('../utils');
+const { logSuccessMessage } = require('./utils');
+
+const SIGNATURE_KEYS_CHECK_ENABLED = false;
+
+function verifyData() {
+  console.info('Verifying dAPI Management data locally...');
```
pieterdub, line 21: It is not really helpful to see `locally` in the CI. Could add confusion.
```suggestion
  console.info('Verifying dAPI Management data...');
```

scripts/verification/verify-dapi-pricing-data.js
```diff
@@ -0,0 +1,37 @@
...
+const forEach = require('lodash/forEach');
+const currentHashData = require('../../data/dapi-pricing-merkle-tree-root/current-hash.json');
+const signerData = require('../../data/dapi-pricing-merkle-tree-root/hash-signers.json');
+const { createDapiPricingMerkleTree, deriveTreeHash, getDapiPricingHashType } = require('../utils');
+const { logSuccessMessage } = require('./utils');
+
+const SIGNATURE_KEYS_CHECK_ENABLED = false;
+
+function verifyData() {
+  console.info('Verifying dAPI Pricing data locally...');
```
pieterdub, line 12: Same here

scripts/verification/verify-signed-api-url-data.js
```diff
@@ -0,0 +1,57 @@
...
+const { getOisTitlesWithAirnodeAddress } = require('@api3/api-integrations');
+const currentHashData = require('../../data/signed-api-url-merkle-tree-root/current-hash.json');
+const signerData = require('../../data/signed-api-url-merkle-tree-root/hash-signers.json');
+const { createSignedApiUrlMerkleTree, deriveTreeHash, getSignedApiUrlHashType } = require('../utils');
+const { logSuccessMessage } = require('./utils');
+
+const SIGNATURE_KEYS_CHECK_ENABLED = false;
+
+function verifyData() {
+  console.info('Verifying Signed API URL data locally...');
```
pieterdub, line 15: Ditto

scripts/generate-dapi-management-mt.js
```diff
@@ -1,37 +1,18 @@
 const fs = require('fs');
 const { ethers } = require('ethers');
-const { deriveBeaconSetId, createDapiManagementMerkleTree } = require('./utils');
-const { getAirnodeAddressByAlias, deriveDataFeedId } = require('@api3/api-integrations');
-const { deriveWalletPathFromSponsorAddress } = require('@api3/airnode-node/dist/src/evm');
+const { createDapiManagementMerkleTree, deriveDataFeedId, deriveSponsorWalletAddress } = require('./utils');
 const { execSync } = require('child_process');
+const dapis = require('../data/dapis.json');
```
pieterdub, line 5: Do we want to include `../data/dapis.json` in the .gitignore file 🤔 

Will there be similar `dapis.json` files for the other trees? If so perhaps it should be moved to the `dapi-management` data folder

scripts/verification/verify-dapi-management-data.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // Verify signatures
+  const hashToSign = deriveTreeHash(getDapiManagementHashType(), currentHashData.hash, currentHashData.timestamp);
+  forEach(currentHashData.signatures, (signature, address) => {
+    assert.equal(address, ethers.utils.verifyMessage(hashToSign, signature), 'Expected all signatures to be valid');
+  });
+
+  // Verify merkle tree values
+  currentHashData.merkleTreeValues.forEach((values) => {
```
pieterdub, line 50:
```suggestion
  currentHashData.merkleTreeValues.forEach((value) => {
```

scripts/verification/verify-signed-api-url-data.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  // Verify signatures
+  const hashToSign = deriveTreeHash(getSignedApiUrlHashType(), currentHashData.hash, currentHashData.timestamp);
+  forEach(currentHashData.signatures, (signature, address) => {
+    assert.equal(address, ethers.utils.verifyMessage(hashToSign, signature), 'Expected all signatures to be valid');
+  });
+
+  // Verify that we've got entries in the @api3/api-integrations package for all airnode addresses
+  currentHashData.merkleTreeValues.forEach((values) => {
```
pieterdub, line 44:
```suggestion
  currentHashData.merkleTreeValues.forEach((value) => {
```

pieterdub (changes requested):
Few minor comments, happy to discuss if they are to petty 😄

scripts/generate-dapi-management-mt.js
mcoetzee: > Do we want to include ../data/dapis.json in the .gitignore file

The file was added as part of this [PR](https://github.com/api3dao/dapi-management/pull/79) (script that generates dAPI management mt data), and we definitely don't want to delete it 🙂

scripts/verification/verify-dapi-management-data.js
mcoetzee: We are dealing with an array of arrays, so imo it's not ideal to give an array a singular variable name.

scripts/verification/verify-dapi-management-data.js
pieterdub: I understand, perhaps `Verifying dAPI Management local data...` but happy if you keep it as is 👍 

Thanks for the explanation

scripts/verification/verify-dapi-management-data.js
pieterdub: True, the only reason I like the singular form is that this particular array pertains to a single tree value. But I suppose each tree entry has multiple values so happy if you keep it as is

pieterdub (approved):
LGTM thanks for a few explanations

PR #94, opened 2023-12-11
Fix gh-pages action when it has empty changes

pieterdub:
Update gh-action to not try commit and push if there are empty changes. 

image


mcoetzee (approved):
👍

PR #95, opened 2023-12-11
Update github-pages workflow by separating out running script and git commit/push 

pieterdub:
Changes has been made to the `github-pages.yml` to facilitate only commit and push when there are actual files that need to be pushed to the branch.

Workflow follows like:
1. checkout gh-pages
2. checkout required files
3. install packages
4. run the script <- This step will always run to create all the possible changes before we decide to push
5. if there are changes to the data directory commit and push

There was also a mismatch of format with main and gh-pages branch. The `dapi-pricing` split script runs `yarn format` and without the config we have in main branch this created discrepancies with the data folder but after the formatter runs with the default config there are no changes detected.

Format currently on `gh-pages`
image

Format currently on `main`
image

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -12,9 +12,10 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
 
 Due to the size of `dapi-pricing` merkle tree values we need to split the values in separate files.
 The workflow will:
+
 1. Checkout the required files including the script that needs to split the values
 2. Install packages that is required in the script
-3. Run the script that splits the values in separate files 
+3. Run the script that splits the values in separate files
 4. Commit and push if there are changes to the data directory
 
 When changes to the `gh-pages` branch are picked up, it will automatically start a new deploy a new update to github pages which can be accessed on:
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,6 +8,15 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
 - **Job:** The workflow consists of a single job named `deploy`.
 - **Runner:** The job runs on the latest version of Ubuntu.
 
+## dApi Pricing
+
+Due to the size of `dapi-pricing` merkle tree values we need to split the values in separate files.
+The workflow will:
+1. Checkout the required files including the script that needs to split the values
+2. Install packages that is required in the script
+3. Run the script that splits the values in separate files 
+4. Commit and push if there are changes to the data directory
+
 When changes to the `gh-pages` branch are picked up, it will automatically start a new deploy a new update to github pages which can be accessed on:
 
 ```
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,17 +8,15 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
 - **Job:** The workflow consists of a single job named `deploy`.
 - **Runner:** The job runs on the latest version of Ubuntu.
 
-## dApi Pricing
-
-Due to the size of `dapi-pricing` merkle tree values we need to split the values in separate files.
+Due to the size of `dapi-pricing` merkle tree values we need to split the values into separate files.
 The workflow will:
 
 1. Checkout the required files including the script that needs to split the values
-2. Install packages that is required in the script
-3. Run the script that splits the values in separate files
+2. Install packages in order for the split script to run
+3. Run the script that splits the values into separate files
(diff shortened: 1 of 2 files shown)


.github/workflows/github-pages.yml
```diff
@@ -32,21 +32,24 @@ jobs:
 
     - name: Install dependencies
       run: |
-        if [ -n "$(git status data scripts/split-dapi-pricing.js --porcelain)" ]; then
-          yarn install --frozen-lockfile
-        fi
+        yarn install --frozen-lockfile
```
mcoetzee, line 35: Minor: This can run on a single line now: 
```yml
run: yarn install --frozen-lockfile
```

.github/workflows/github-pages.yml
```diff
@@ -32,21 +32,24 @@ jobs:
...
+
+    - name: Run split script
+      run: |
+        echo "Running dapi pricing split script"
```
mcoetzee, line 39: Minor: This echo is a bit unnecessary now that we have a separate step for the script, so I think:
```yml
run: node scripts/split-dapi-pricing.js
```

.github/workflows/github-pages.yml
```diff
@@ -32,21 +32,24 @@ jobs:
...
       run: |
-        if [ -n "$(git status data scripts/split-dapi-pricing.js --porcelain)" ]; then
+        if [ -n "$(git status data --porcelain)" ]; then
           echo "Changes detected in gh-pages branch."
           git config user.name github-actions
           git config user.email github-actions@github.com
-          echo "Running dapi pricing split script"
-          node scripts/split-dapi-pricing.js
           git add data
+          echo "Commiting file"
```
mcoetzee, line 49: Minor: We can probably get rid of this echo now?

mcoetzee (changes requested):
Can we update the PR title please. This PR fixes the GH pages workflow, which isn't evident from the current title: `Include prettier config in GH action to align with main`

Edit: And can you also link to the issue https://github.com/api3dao/dapi-management/issues/63

.github/workflows/README.md
```diff
@@ -8,6 +8,16 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
 - **Job:** The workflow consists of a single job named `deploy`.
 - **Runner:** The job runs on the latest version of Ubuntu.
 
+## dApi Pricing
```
mcoetzee, line 11: I don't think this should be its own section. Reckon you can remove this heading and the below can form part of the details section.

.github/workflows/README.md
```diff
@@ -8,6 +8,16 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
...
+
+Due to the size of `dapi-pricing` merkle tree values we need to split the values in separate files.
```
mcoetzee, line 13: *`... split the values into ...`

.github/workflows/README.md
```diff
@@ -8,6 +8,16 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
...
+The workflow will:
+
+1. Checkout the required files including the script that needs to split the values
+2. Install packages that is required in the script
```
mcoetzee, line 17: Minor: Sounds like we are only installing packages that are required in the script. So I think: `Install packages in order for the split script to run`

.github/workflows/README.md
```diff
@@ -8,6 +8,16 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
...
+3. Run the script that splits the values in separate files
+4. Commit and push if there are changes to the data directory
+
 When changes to the `gh-pages` branch are picked up, it will automatically start a new deploy a new update to github pages which can be accessed on:
```
mcoetzee, line 21: This is a bit off: `... start a new deploy a new update to ...`

.github/workflows/README.md
```diff
@@ -8,6 +8,16 @@ This GitHub Actions workflow detects changes to the data folder on the main bran
...
 - **Runner:** The job runs on the latest version of Ubuntu.
 
+## dApi Pricing
+
+Due to the size of `dapi-pricing` merkle tree values we need to split the values in separate files.
+The workflow will:
+
+1. Checkout the required files including the script that needs to split the values
+2. Install packages that is required in the script
+3. Run the script that splits the values in separate files
```
mcoetzee, line 18: *`... the values into ...`

mcoetzee (approved):
Nice one 👍

PR #106, opened 2023-12-13
Add script to generate dapi pricing merkle tree

vponline:
Closes [Add a dAPI pricing Merkle tree update script to dapi-management using the old pricing model and the new database as the data reference#519](https://github.com/api3dao/tasks/issues/519)

Note that we don't have signed data or gas data collectors deployed yet so testing this with real data is not yet possible.


src/database/database.js
```diff
@@ -0,0 +1,13 @@
+require('dotenv').config();
```
mcoetzee, line 1: Will we be using this db client anywhere else in the foreseeable future? I have some doubts whether it makes sense to have `src` folder in this repo, so I'm thinking you could declare this locally for now. 

Wrt the `src` folder (as a fyi), my current thinking is that we could remove it at some point and have a root `lib` folder where we put common code that both the `scripts` and `frontend` code could use. This would also involve converting the scripts to Typescript (where we could use `ts-node` to run them).

scripts/generate-dapi-pricing-mt.js
```diff
@@ -0,0 +1,31 @@
...
+
+const MT_INPUT_PATH = './scripts/dapi-pricing-mt-input.json';
+const MT_OUTPUT_PATH = './data/dapi-pricing-merkle-tree-root/current-hash.json';
+
+async function generateDapiPricingMT() {
+  const merkleTreeValues = JSON.parse(fs.readFileSync(MT_INPUT_PATH, 'utf8'));
+
+  const tree = createDapiPricingMerkleTree(merkleTreeValues);
+
+  const dapiPricingMT = {
```
mcoetzee, line 13: I think it makes sense to short circuit this script if the hashes match. Along the lines of:
```js
if (tree.hash === currentData.hash) {
  console.info('Current hash file is up to date');
  return;
}
```

mcoetzee (commented):
I will need someone more familiar in the pricing space to review the queries, calculation etc of the merkle tree values. 
Also, I'm not quite sure I follow the workflow here. Why do we need two scrips and to store a .gitignored file?

vponline:
>Also, I'm not quite sure I follow the workflow here. Why do we need two scrips and to store a .gitignored file?

To fix for Burak's request that the MT generation script to always generate the same tree, I split the script into two:
1. Generate the inputs from the database (live data) so the results will change on each run. This file is gitignored.
2. Load the inputs file into the MT generation script and generate the tree. This will produce the same tree each time unless script 1 is ran again.

bbenligiray:
Why can't this happen in a single step?

vponline:
hmm don't think I was following initially and thought the Merkle Tree parameters would need to be calculated separately to avoid overriding them on each run on the Merkle Tree generation script. 

Unless we hardcode the start date and that would need to be set manually be the user or defined as a script input 🤔. That would work in one script.

>Can you extract all parameters to a JSON outside of the script please?

I guess you meant these script input parameters instead of the merkle tree parameters 😅, sorry about that. I'll fix

bbenligiray:
So the intended workflow is, for example Jiri comes in at the end of the month and opens a PR. He updates the hardcoded parameters in a JSON file (start and end dates of the data to be used, chain multipliers, etc.) and runs the script (the script that this issue is about) to update the current Merkle tree and he signs it. Other people come in, check the hardcoded parameters for any changes, run the same script to verify that the Merkle tree is correct, and they also sign it. Then it gets registered on-chain.

vponline:
> So the intended workflow is, for example Jiri comes in at the end of the month and opens a PR. He updates the hardcoded parameters in a JSON file (start and end dates of the data to be used, chain multipliers, etc.) and runs the script (the script that this issue is about) to update the current Merkle tree and he signs it. Other people come in, check the hardcoded parameters for any changes, run the same script to verify that the Merkle tree is correct (maybe the script should take a parameter to know if it needs to update or verify the current Merkle tree), and they also sign it. Then it gets registered on-chain.

Thanks for clarifying, fixed now.

scripts/generate-dapi-pricing-mt.js
```diff
@@ -0,0 +1,371 @@
...
+}
+
+function checkDeviation(currentValue, nextValue, threshold) {
+  const absoluteDifference = Math.abs(currentValue - nextValue);
+  const percentageDifference = (absoluteDifference / Math.max(Math.abs(currentValue), Math.abs(nextValue))) * 100;
+
+  return percentageDifference > threshold;
+}
+
+// This function is copied from utils, but removes the nodary restriction since we need the BeaconSet consituent dataFeedIds as well
```
aquarat, line 61:
```suggestion
// This function is copied from utils, but removes the nodary restriction since we need the BeaconSet constituent dataFeedIds as well
```

aquarat (commented):
I don't see any specific issues with this, it's obviously quite complex.

I tried to run it locally but wasn't able to due to being unable to connect to the DB (I am able to connect using `psql` directly)

I might be losing it but the README doesn't explicitly say how to run the command included? I tried to run it with `node scripts/generate-dapi-pricing-mt.js`

bbenligiray:
Consider having @RiVal-cz test it

aquarat:
Just FYI I ran into this connection issue again and I solved it by modifying the code like this:
```typescript
const creds = {
  connectionString: 'postgresql://postgres:password@db-url/postgres?schema=public',
  ssl: {
    rejectUnauthorized: false,
  }
};

// Set the database credentials
const client = new Client(creds);
```

This can probably be solved by generating a certificate for the server.

vponline:
I've added the ssl field to the DB client to enable running with the production DB for now. We'll still need real Signed API data to be able to test this with live data.

scripts/generate-dapi-pricing-mt.js
```diff
@@ -0,0 +1,374 @@
...
+  const adjustedGasPrice = ethers.BigNumber.from(Math.ceil(gasPriceMultiplier.mul(averageGasPrice).toNumber()));
+
+  const chainSingleUpdateGasCost = chainSingleUpdateGasCosts[chainId] ?? {
+    fixedGasCost: defaultSingleUpdateGasCost,
+  };
+
+  switch (chainId) {
+    case '10' /* optimism*/:
+    case '420' /* optimism-testnet*/:
+    case '5000' /* mantle*/:
```
aquarat, line 91: You've already noted this on Slack I think but Mantle pricing is different :(

aquarat (approved):
LGTM 👌 
Best way to test/review it is try and use it a few times with actual signed data 😄

PR #107, opened 2023-12-13
Remove Tree Values and Tree Diff tabs from dAPI Pricing MT page

### LINKED ISSUE #82: Temporarily hide pricing MT table from dapi-management FE until Pricing v2 is finalized
Due to the file size, the pricing MT table containing the values should be temporarily hidden from the dapi-management FE until Pricing v2 method is finalised for Phase 2.1. It should be noted that Pricing v2 is not a launch blocker for Phase 2.1.

However, since the Pricing MT is currently generated with a script from the Pricing v1 method (*script related to https://github.com/api3dao/tasks/issues/399), the dapi-management FE should still have a button to sign the root of these (Pricing v1) generated pricing MT.

Hence, the dAPI Pricing Merkle Tree FE page, can only retain for now the root, the sign button and the signers.

- [ ] Hide MT table with values. 
- [ ] Remove price MT from FE and hide caching for other trees since they are small enough.
- [ ] Retain for now the root, the sign button and the signers
- [ ] Update README.md to document clearly how signers should use the script* to generate the root locally. It should include key steps like cloning the repo, running the script* locally, and run git diff to ensure there is no differences before signing the root.

pieterdub:
Closes #82 

Hide data using a constant flag that can be set if need be and include a placeholder text explaining that the data has been hidden intentional.

image

### DIFF
--- a/frontend/pages/merkle-trees/dapi-pricing.tsx
+++ b/frontend/pages/merkle-trees/dapi-pricing.tsx
@@ -1,4 +1,3 @@
-import { z } from 'zod';
 import RootLayout from '~/components/root-layout';
 import { TreeStatusBadge, TreeRootBadge, SignatureTable, SignRootButton } from '~/components/merkle-tree-elements';
 import { validateTreeRootSignatures } from '~/lib/merkle-tree-utils';
@@ -6,19 +5,12 @@ import { InferGetServerSidePropsType } from 'next';
 import { useTreeSigner } from '~/components/merkle-tree-elements/use-tree-signer';
 import { readSignerDataFrom, readTreeDataFrom } from '~/lib/server/file-utils';
 
-const merkleTreeSchema = z.object({
-  timestamp: z.number(),
-  hash: z.string(),
-  signatures: z.record(z.string()),
-});
-
 export async function getServerSideProps() {
   const subfolder = 'dapi-pricing-merkle-tree-root';
 
   const { data: currentTree } = readTreeDataFrom({
     subfolder,
     file: 'current-hash.json',
-    schema: merkleTreeSchema,
   });
 
   const { data: signers } = readSignerDataFrom(subfolder);
--- a/frontend/pages/merkle-trees/dapi-pricing.tsx
+++ b/frontend/pages/merkle-trees/dapi-pricing.tsx
@@ -57,7 +57,7 @@ export default function DapiPricingTree(props: Props) {
       <div className="mb-10">
         <SignatureTable signers={signers} signatures={signatures} />
       </div>
-      <div className="flex justify-center text-lg text-gray-400">
+      <div className="flex justify-center text-gray-400">
         dAPI Pricing merkle tree data has been hidden due to its size
       </div>
     </RootLayout>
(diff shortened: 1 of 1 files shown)


frontend/pages/merkle-trees/dapi-pricing.tsx
```diff
@@ -93,57 +95,63 @@ export default function DapiPricingTree(props: Props) {
...
-              <TableBody>
-                {currentTree.merkleTreeValues.slice(0, 100).map((pricing, i) => (
-                  <TableRow key={i}>
-                    <TableCell>{ethers.utils.parseBytes32String(pricing[0])}</TableCell>
-                    <TableCell>{getChainAlias(pricing[1])}</TableCell>
-                    <TableCell>{formatUpdateParams(pricing[2])}</TableCell>
-                    <TableCell>{formatDuration(pricing[3])}</TableCell>
-                    <TableCell>{convertPrice(pricing[4])}</TableCell>
+      {SHOW_DAPI_PRICING_DATA ? (
+        <Tabs defaultValue="0">
```
mcoetzee, line 99: We don't want to maintain all this code so lets remove it all. This includes the unnecessary page props etc (we don't need to call `getMerkleTreeServerSideProps`). We can get it back from Git if we need it 👍

frontend/pages/merkle-trees/dapi-pricing.tsx
```diff
@@ -93,57 +95,63 @@ export default function DapiPricingTree(props: Props) {
...
+                </TableBody>
+              </Table>
+            )}
+          </TabsContent>
+          <TabsContent value="1" forceMount>
+            <TreeDiff diffResult={props.diffResult} diffMode={diffMode} raw={showRawValues} />
+          </TabsContent>
+        </Tabs>
+      ) : (
+        <div className="mt-36 flex w-full justify-center text-lg text-gray-400">
```
mcoetzee, line 151: Minor: Hmm I don't think this text needs to be so prominent. I would probably remove the `mt-36` and the `text-lg` (and the `w-full` is not needed).

mcoetzee (changes requested):
Can update the title to something like "Remove Tree Values and Tree Diff tabs from dAPI Pricing MT page" please 🙏

frontend/pages/merkle-trees/dapi-pricing.tsx
```diff
@@ -1,66 +1,37 @@
...
 });
 
-export async function getServerSideProps(context: GetServerSidePropsContext) {
-  const showRawValues = context.query.raw === 'true';
+export async function getServerSideProps() {
+  const subfolder = 'dapi-pricing-merkle-tree-root';
 
-  const { currentTree, signers, diffResult } = await getMerkleTreeServerSideProps({
-    subfolder: 'dapi-pricing-merkle-tree-root',
+  const { data: currentTree } = readTreeDataFrom({
```
mcoetzee, line 18: We can actually remove this custom `merkleTreeSchema` now as well (it was declared to have more fine grained validation/types for the `merkleTreeValues` which we don't need now 👍

mcoetzee (approved):
One last minor comment, so I'm approving so long 👍

PR #127, opened 2024-01-02
Update supported chains generation

### LINKED ISSUE #126: Improve exporting supported chains
**Acknowledgements**
- Exporting supported chains should be readible

**To-Do**
- Replace chain ids with chain aliases

hiletmis:
Closes #126

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,10 @@
 import dapis from '../data/dapis.json';
-import supportedChainIds from './generated/supported-chains.json';
+import supportedChains from './generated/supported-chains.json';
 import * as api3Chains from '@api3/chains';
 import * as api3Integrations from '@api3/api-integrations';
 
-function supportedChains() {
-  return supportedChainIds
+function supportedChainIds() {
+  return supportedChains
     .map((alias) => {
       const chain = api3Chains.CHAINS.find((chain) => chain.alias === alias);
       if (!chain) {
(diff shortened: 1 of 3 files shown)


hiletmis:
@mcoetzee `dapi-management` package calculates supported chains and write them into `supported-chains.json` file as chain ids array.

To simplify review processes chain ids replaced with chain aliases using `chains` package.

src/index.ts
```diff
@@ -3,4 +3,16 @@ import supportedChains from './generated/supported-chains.json';
 import * as api3Chains from '@api3/chains';
 import * as api3Integrations from '@api3/api-integrations';
 
-export { dapis, supportedChains, api3Chains, api3Integrations };
+function supportedChainIds() {
```
mcoetzee, line 6: Minor: Should we perhaps call this function `getSupportedChainIds`?

mcoetzee (approved):
Just got one minor comment, so I'm approving so long 👍

src/index.ts
hiletmis: Thanks updated function name as suggested.

hiletmis: But changing the name may break previously usages, idk just wondering if its the case

mcoetzee: Previous usages would already be affected though, because the `supportedChains` array was changed to a function.

